PR #565, opened 2013-06-21
bring back :command_executed + configurable anonymize attributes

piavlo:
a desparate attempt to bring back :command_executed  :)


portertech:
@piavlo I'm going to close this, as much has changed in Sensu Core since the creation of this PR. Many Sensu users still use command token substitution and attribute redaction as security features, which this feature would effectively bypass. Sensu environment variables are a better approach, but this is still the case.

PR #632, opened 2013-09-30
Ensure clients are removed from the list before removing them

portertech:


sa-shukla:
Yes. I have setup Redis HA and whenever Redis failover happens, sensu-api crashes. Though sensu-server is able to reconnect to the redis-master.

I am running sensu version 0.19.1

Seems this issue is open since couple of years and the maintenance team should look into this more seriously. 

Regards,
Sagar

PR #697, opened 2014-02-12
Fix check extensions

christophergeers:
This change addresses an issue preventing extension checks from working by validating check configurations contain either a :command or :extension key.

It also allows extension checks to set the extension to be run based on check[:extension] rather than check[:name].  This allows for one extension to be scheduled many times with unique names and (potentially) different parameters.

I've tested this briefly on 0.12.1 with much success and happiness.


portertech:
This feature is now in master, along with support for the standard extension behaviour https://github.com/sensu/sensu/pull/889

PR #707, opened 2014-04-03
Make sensu server not merge check data when the check is stand_alone

solarkennedy:
I just want to spark the conversation, if approved I'll try to add tests. (I don't know much ruby)

I have this situation with @bobtfish where we have lots of standalone checks in a multi-tenant environment. (also, the default configuration when the puppet module is uses, `standalone => true`)

But, in the situation where a client sends in a check result from the standalone check, and it is "missing" a critical piece of metadata, say `pagerduty_api_key`, the _server_ gleefully fills in this data with this merge.

In my particular situation this results in lots of pages going to a the team that owns the sensu _server_ (Ops), and not to the team that owns the client (not Ops).

Can we make the sensu server _not_ perform this merge when the check is standalone? It seems like a standalone check should have all settings it needs to do its job, and the server shouldn't be filling in the blanks just because it also happens to have the same check defined for itself, with different settings.

Or, let me know if this is not the design you want sensu to take, and I can come up with a different workaround (prepend all checks with the hostname so the check names never merge?)

### DIFF
--- a/lib/sensu/server.rb
+++ b/lib/sensu/server.rb
@@ -413,7 +413,7 @@ module Sensu
         unless client_json.nil?
           client = Oj.load(client_json)
           check = case
-          when @settings.check_exists?(result[:check][:name])
+          when @settings.check_exists?(result[:check][:name]) && !result[:check][:standalone]
             @settings[:checks][result[:check][:name]].merge(result[:check])
           else
             result[:check]


srodrig0209:
This doesn't appear to be working for me. All clients/servers are running 0.16.0. I upgraded the sensu servers first (from 0.12.6), flushed redis, then upgraded the clients. 

On the sensu server:
```json
{
  "checks": {
    "check_load": {
      "command": "/etc/sensu/plugins/check-load.rb -w 10,10,10 -c 12,12,12",
      "handlers": [
        "default",
        "sns",
        "logevent",
        "graphite_event"
      ],
      "standalone": true,
      "interval": 90,
      "refresh": 300,
      "occurrences": 2
    }
  }
}
```

On another node:
```json
{
  "checks": {
    "check_load": {
      "command": "/etc/sensu/plugins/check-load.rb -w 12,12,12 -c 14,14,14",
      "handlers": [
        "default",
        "sns",
        "logevent",
        "graphite_event"
      ],
      "standalone": true,
      "interval": 90,
      "refresh": 300,
      "occurrences": 2
    }
  }
}
```

What's shown on Uchiwa for the latter node:
```
command     /etc/sensu/plugins/check-load.rb -w 10,10,10 -c 12,12,12
```

solarkennedy:
Yes, in the API (and uchiwa) the check definition is for whatever the server is.

At event *processing* time, the server will use the standalone-version that the client provided.

srodrig0209:
:disappointed: Sigh. Yeah, that's what I thought. I knew the checks were being executed properly, but if you think of it from a new user perspective, this doesn't make any sense.

It should be noted somewhere that "hey, that command definition you're looking at is false because you decided to use 'standalone' ". Just my 2 cents.

solarkennedy:
I agree that it is confusing, but it isn't "false".
How about submit a PR on the FAQ page in the docs repo?

The API is returning the information requested, which is "what server-side checks are defined", even if that api endpoint is just "/checks" :(

srodrig0209:
Well, maybe false was a poor word choice. One would think that if standalone then the client is the source of truth, otherwise the server knows the check definition. Yeah, I'll draft up something for the docs.

PR #752, opened 2014-06-19
adding support for length prefixed json on the client socket

praddy:
As @portertech & I agreed on IRC, I am adding support for length prefixed JSON to the client socket while keeping the existing behavior as it is.

Thanks,
praddy


simondean:
@praddy sounds like length prefixed events would be a robust approach.  

BTW the format of the length header you've used looks good: `/^(\d+)\n/n`

Issue #815, opened 2014-11-03
sensu-api just quits once in a while (after sensu-server restart usually)


FlorinAndrei:
I've seen this today on version 0.16.0, but it's been happening ever since I've started using Sensu, a few versions back.

Basically, sensu-api just quits. This seems to happen more often if I restart sensu-server, so I acquired the habit of restarting sensu-api too a few seconds after restarting sensu-server - that seems to help somewhat.

These are the messages I've seen in the logs today:

```
{"timestamp":"2014-11-03T11:02:00.103424-0800","level":"info","message":"GET /clients","remote_address":"127.0.0.1","user_agent":null,"request_method":"GET","request_uri":"/clients","request_body":""}
{"timestamp":"2014-11-03T11:02:00.107152-0800","level":"info","message":"GET /events","remote_address":"127.0.0.1","user_agent":null,"request_method":"GET","request_uri":"/events","request_body":""}
{"timestamp":"2014-11-03T11:02:10.108778-0800","level":"info","message":"GET /info","remote_address":"127.0.0.1","user_agent":null,"request_method":"GET","request_uri":"/info","request_body":""}
{"timestamp":"2014-11-03T11:02:10.109866-0800","level":"fatal","message":"transport connection error","error":"rabbitmq channel closed"}
{"timestamp":"2014-11-03T11:02:10.109973-0800","level":"warn","message":"stopping"}
{"timestamp":"2014-11-03T11:02:10.610888-0800","level":"warn","message":"stopping reactor"}
```

I cannot be sure if these same messages were printed during previous crashes, I was too busy to look at the logs until today, sorry.

If I do "service sensu-api restart", it resumes working without fuss.

I'm also using whatever is the latest Uchiwa and RabbitMQ in the repos, all on Ubuntu 14.04 fully updated. A couple dozen clients spread through the cloud in various geographies, usually running same Sensu version as the server.

srodrig0209:
I also just experienced this. I terminated a secondary dev instance and saw the following:

```
=INFO REPORT==== 14-Jan-2015::11:37:48 ===
accepting AMQP connection <0.13516.18>

=WARNING REPORT==== 14-Jan-2015::11:37:48 ===
closing AMQP connection <0.13516.18> :
connection_closed_abruptly

=INFO REPORT==== 14-Jan-2015::11:37:50 ===
Statistics database started.

=INFO REPORT==== 14-Jan-2015::11:37:50 ===
[...]

Issue #599, opened 2013-08-16
sensu-server not reconnecting after redis crash and failover


bencaron:
Just had a issue with sensu-server that did not reconnected to redis server after a crash of the redis server.

We are using a HA redis setup as described here:
https://blog.theroux.ca/sensu/high-availability-sensu/

The master redis server (01) had a kernel panic. The failover worked as expected, with the master switching to 02 and HAProxy marking the redis 01 as down and 02 as up.

Normally, sensu should have reconnected automatically:

https://github.com/portertech/em-redis/blob/unified/lib/em-redis/redis_protocol.rb#L490

But the events started piling up into rabbitmq for a good 15 minutes until the sensu servers (we have 2) started processing them again. CPU load on the sensu servers turned to almost none and almost nothing got logged in the mean time:

In the logs:

```
{"timestamp":"2013-08-16T09:34:44.885847-0400","level":"info","message":"only handling every 30 occurrences: my-server-name/check_swap"}
{"timestamp":"2013-08-16T09:50:41.209103-0400","level":"warn","message":"reconnecting to redis"}
```

The part missing from the redis HA tutorial is how the HAProxy is configured:

```
backend bk_redis
  log     global
  mode    tcp
  option  log-health-checks
  option  tcplog
  option  httpchk GET /
  balance leastconn
  timeout connect 10s
  timeout queue 5s
  timeout server 3600s
  server redis-01 xx.yy.zz.aa:6379 check port 12345
  server redis-02 xx.yy.zz.bb:6379 check port 12345
```

Maybe the socket got stuck between sensu-server and HAProxy, but since the redis-01 was in kernel panic, it was not a socket left in a incomplete state.

We had a similar but not-quite-the-same (and not as throughly examined) incident this week. I'll update this issue with more details if the problem happens again.

I know that it's not absolutely obvious to figure what went wrong here, but thanks for tips on what could have went wrong.

dblessing:
I just experienced the same thing today. I was baffled when sensu-api and sensu-server on all my nodes just stopped. Same log output as noted by @johntdyer

sa-shukla:
Yes. I have setup Redis HA and whenever Redis failover happens, sensu-api crashes. Though sensu-server is able to reconnect to the redis-master.

I am running sensu version 0.19.1

Seems this issue is open since couple of years and the maintenance team should look into this more seriously.

Regards,
Sagar

portertech:
[...]

PR #845, opened 2015-01-16
[Feedback Requested] Attempt to reconnect if a RabbitMQ or Redis error occurs

dblessing:
This fix works in my environment and I believe it's a good/possible fix for #815 and #599. See my comment https://github.com/sensu/sensu/issues/815#issuecomment-70260550 fully explaining the cycle. Summary: In an HA environment it's possible that something like Redis will fail and while the load balancer swaps over to the new Redis master one of 2 things could happen:

1. The load balancer swaps to the new Redis before it has fully achieved master status. Thus, Sensu will receive a "READONLY slave" error. Let's assume that the error will eventually resolve itself and keep trying.
1. When Sensu disconnects from Redis it also unsubscribes from the RabbitMQ keepalive and results queue. As a result, the `sensu-api` process will try to reconnect to RabbitMQ, and when those queues are not available it will bomb out and kill the process. Again, let's assume that the situation is not fatal and keep trying. 

Even in a non-HA environment I don't see the harm in throwing an error (logging it) and retrying. The situation may not correct itself but then it will just cycle until someone looks at the logs and see it's not working. Or better yet, they're monitoring Sensu API health and will notice it's not connected to Redis and/or RabbitMQ.

Do you think this might be an acceptable answer to these issues? If not, what else can we do? As it stands Sensu is completely vulnerable to Redis failures even when Redis is set up in a cluster.

Thanks for your consideration.

### DIFF
--- a/lib/sensu/daemon.rb
+++ b/lib/sensu/daemon.rb
@@ -132,10 +132,11 @@ module Sensu
       Transport.logger = @logger
       @transport = Transport.connect(transport_name, transport_settings)
       @transport.on_error do |error|
-        @logger.fatal('transport connection error', {
+        @logger.error('transport connection error', {
           :error => error.to_s
         })
-        stop
+        pause
+        setup_transport
       end
       @transport.before_reconnect do
         unless testing?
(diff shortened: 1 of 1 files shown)


portertech:
Master now has support for:

```
{
    "transport": {
        "reconnect_on_error": "true"
    },
    "redis": {
        "reconnect_on_error": "true"
    }
}
```

Will be part of the 0.17.0 stable build.

portertech:
The new functionality should handle the failure cases you've stated above :)

jjshoe:
@portertech - Will the transport piece close out https://github.com/sensu/sensu/issues/815

portertech:
@jjshoe the option does fix #815, but it's not a default setting (yet).

PR #847, opened 2015-01-20
Add support for a subdue 'schedule' to give more flexibility

mrichar1:
This is a first-stab at a subdue 'schedule' which would give more flexibility in setting subdue times over the current subdue code.

The idea is that in the config you can set a subdue schedule which allows multiple subdue periods across a day - using day of week number as key, and a list of lists for begin and end pairs. For example to subdue checks completely on Saturday and Sunday, outside of 9-5 Monday, outside 10-4 on a Friday, and not at all for the rest of the week, you would add the following config:

``` json
"subdue": {
  "schedule": {
    "0": [],
    "1": [
          ["00:00", "'09:00"],
          ["17:00", "23:59:59"]
        ],
    "5": [
          ["00:00", "10:00"],
          ["16:00", "23:59:59"]
        ],
    "6": []
  }
}
```

I've tested the code, and it seems to do what I expect, but I would appreciate a review!  I'm also happy for any changes to the config structure (e.g. using day names instead of numbers) if preferred. I've added this functionality alongside the existing subdue options, so that it would be backwards compatible, but I've not thought much/tested how this would play out if both old and new options were set.

Mostly I'm just keen to know if this is a useful change!

### DIFF
--- a/lib/sensu/server.rb
+++ b/lib/sensu/server.rb
@@ -48,6 +48,25 @@ module Sensu
 
     def action_subdued?(condition)
       subdued = false
+      if condition.has_key?(:schedule)
+        now = Time.now
+        day_of_week = now.wday
+        timepairs = condition[:schedule][day_of_week]
+        # If timepairs isn't an array, then there was no entry for today
+        if timepairs.kind_of?(Array)
+           if timepairs.length == 0
+             # No times given - subdue for the whole day
+             subdued = true
+           else
+             timepairs.each do |timepair|
+              if now.between?(Time.parse(timepair[0]), Time.parse(timepair[1]))
+                subdued = true
+                break
+              end
+            end
+          end
+        end
+      end
       if condition.has_key?(:begin) && condition.has_key?(:end)
         begin_time = Time.parse(condition[:begin])
         end_time = Time.parse(condition[:end])
--- a/lib/sensu/server.rb
+++ b/lib/sensu/server.rb
@@ -50,8 +50,7 @@ module Sensu
       subdued = false
       if condition.has_key?(:schedule)
         now = Time.now
-        day_of_week = now.wday
-        timepairs = condition[:schedule][day_of_week]
+        timepairs = condition[:schedule][now.wday.to_s]
         # If timepairs isn't an array, then there was no entry for today
         if timepairs.kind_of?(Array)
            if timepairs.length == 0


mrichar1:
Can I bump this PR for comment/review?  Not sure what the travis failure is exactly, but I have a suspicion it's outside of this PR. Please let me know if this isn't the case...

portertech:
@mrichar1 thanks for the pr! :+1: This functionality could certainly be added to the current subdue filtering https://github.com/sensu/sensu/blob/master/lib/sensu/server/filter.rb However, the configuration as it stands is confusing, as the configuration attributes lack context/description.

mrichar1:
Thanks for the review... I agree that without documentation this config is a bit tricky to understand!  It mostly comes down to some hacks which rely on the way the data structures behave in ruby.

I suspect that adding day names instead of numbers would help to clarify this a lot (though I'm not sure what support sensu has for localisation?) - something like:

```json
"subdue": {
  "schedule": {
    "Sunday": [],
    "Monday": [
          ["00:00", "'09:00"],
          ["17:00", "23:59:59"]
        ],
    "Thursday": [
          ["00:00", "10:00"],
          ["16:00", "23:59:59"]
        ],
    "Saturday": []
  }
}
```

A second option would be to make the inner start/end lists dictionaries with named keys, something like:

```json
    "Thursday": [
          {"start": "00:00",
           "end": "10:00"},
          {"start": "16:00",
           "end": "23:59:59"}
        ],
```

I'm not sure if this really adds any clarity though?

The empty list representing 'all day' might be confusing, but then an explicit `["00:00", "23:59:59"]` would also be understood.  I'd be reluctant to try to overload this list with something like `["all"]`.

Any other suggestions for clarity welcome - I'm happy to re-code to support an easier-to-understand format!

derkgort:
Hi,

What is the status of this feature. It's interesting and i'm looking for something like this.
It looks like the build failed because of an error outside of this patch, but somewhere in the sensu code at that moment.

calebhailey:
Tagging this for renewed discussion w/ #1219.

clev2k:
Hi, are there any updates on this?
I've commented on the linked one also as it's really hard to understand which one is the main one with all the cross-referencing.

eklein:
+1 on this, would be fantastic to be able to specify a time window that starts on one day and ends on another. The specific example I'm trying to address is a window that starts on Friday at 7pm and runs until Saturday at 7am. As far as I can tell, there isn't a way to do this other than perhaps filters, but the filter docs have me pretty confused :) I'm needing to create this for a single standalone check that lives on a whole bunch of clients.

calebhailey:
Hi there everyone! We do have an update on this, and it's going to be included in the next release! \o/ 

As part of the [efforts to rethink how subdue and silencing work in Sensu][1] (see [this comment][2] for additional details), subdue definitions will gain [support for a new attribute called `when`][3] that will effectively allow users to subdue check executions  _during_ certain hours, or _outside_ of certain hours, but will allow definitions such as the following: 

~~~ json
"subdue": {
  "at": "publisher",
  "when": "outside",
  "days": [
    "monday",
    "tuesday",
    "wednesday",
    "thursday",
    "friday"
  ],
  "begin": "9AM PT",
  "end": "5PM PT"
}
~~~

We just realized this was previously not documented in a public forum, so I just created a new issue for it here: https://github.com/sensu/sensu/issues/1390. This will be included in the upcoming version 0.26. 

@eklein - this might not solve for the example scenario you provided, but I'd be happy to help you sort through how to write a custom filter. Feel free to hit me up at caleb [at] sensuapp.com. 👍 

Related: https://github.com/sensu/sensu/issues/1219 

[1]: https://github.com/sensu/sensu/issues/1219 
[2]: https://github.com/sensu/sensu/issues/1219#issuecomment-225245100 
[3]: https://github.com/sensu/sensu/issues/1390

portertech:
https://github.com/sensu/sensu/pull/1415 has been merged, implementing this desired functionality (with a different syntax) :+1:

eklein:
Woohoo, thanks @portertech!

PR #848, opened 2015-01-21
Filter attributes need to both be strings to compare correctly.

jmherbst:
While trying to implement a filter for events with `:action=>:resolve` I realized that the value of `:action` in the Sensu Event is a symbol and my filter attributes are in JSON.  The values of [`hash_one[key]` and `hash_two[key]`](https://github.com/sensu/sensu/blob/master/lib/sensu/server.rb#L93) need to be converted to a string before they can accurately be compared.

##### Example Filter JSON: 
 ```JSON
{
  "filters": {
    "do_not_resolve": {
      "attributes": {
        "check": {
          "do_not_resolve": true
        },
        "action": "resolve"
      },
      "negate": true
    }
  }
}
```

##### Example Sensu Event:
```Ruby
{
  :id=>"b024326c-ef4a-4d56-95f9-72332e5a2702", 
  :client=>{
    :name=>"rt4.prod.example.com",
    :address=>"198.11.254.108", 
    :subscriptions=>["base", "rt4.prod.example.com", "testing"], 
    :keepalive=>{
      :thresholds=>{
        :warning=>120, 
        :critical=>180
      }, 
      :handlers=>["default", "pagerduty"], 
      :refresh=>1800
    }, 
    :version=>"0.13.1", 
    :timestamp=>1421791596
  }, 
  :check=>{
    :command=>"monitor_that_pid.sh", 
    :subscribers=>["testing"], 
    :handlers=>["default", "pagerduty"], 
    :refresh=>1800, 
    :interval=>60, 
    :description=>"Monitors Some PID", 
    :do_not_resolve=>true,
    :name=>"monitor_some_pid", 
    :issued=>1421791537, 
    :executed=>1421791537, 
    :duration=>0.013, 
    :output=>"OK", 
    :status=>0, 
    :history=>["2", "2", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "0", "2", "2", "2", "2", "2", "2", "0"]
  }, 
  :occurrences=>6, 
  :action=>:resolve
}
```

I'm pretty sure, at the moment, filtering on any Event `:action` is broken since the actions are symbolized and you can only have a string in the Filter attribute JSON.

I've tested a decent bit, but would appreciate a review or any comments/suggestions!

### DIFF
--- a/lib/sensu/server.rb
+++ b/lib/sensu/server.rb
@@ -90,7 +90,7 @@ module Sensu
     def filter_attributes_match?(hash_one, hash_two)
       hash_one.keys.all? do |key|
         case
-        when hash_one[key] == hash_two[key]
+        when hash_one[key].to_s == hash_two[key].to_s
           true
         when hash_one[key].is_a?(Hash) && hash_two[key].is_a?(Hash)
           filter_attributes_match?(hash_one[key], hash_two[key])


portertech:
I don't think this will work if both values are identical hash objects.

jmherbst:
@portertech -- I think it would work just fine in that case too.  Here is my very limited testing.

Ex:
``` Ruby
> hash_one = {testing: 'test', inside_hash: {new_test: 'also_testing'}}
 => {:testing=>"test", :inside_hash=>{:new_test=>"also_testing"}}
> hash_two = {testing: 'test', inside_hash: {new_test: 'also_testing'}}
 => {:testing=>"test", :inside_hash=>{:new_test=>"also_testing"}}
> hash_three = {testing_failure: 'test', inside_hash: {new_test: 'testing_failure'}}
 => {:testing_failure=>"test", :inside_hash=>{:new_test=>"testing_failure"}}

> hash_one == hash_two
 => true
> hash_one == hash_three
 => false
> hash_one.to_s == hash_two.to_s
 => true
> hash_one.to_s == hash_three.to_s
 => false
```

Could you elaborate a bit more on why you think this will not work?

portertech:
What if the hashes are not sorted? Could end up w/ different strings.

portertech:
```
[1] pry(main)> hash_one = {testing: 'test', inside_hash: {new_test: 'also_testing'}}                                                                 
=> {:testing=>"test", :inside_hash=>{:new_test=>"also_testing"}}
[2] pry(main)> hash_two = {inside_hash: {new_test: 'also_testing'}, testing: 'test'}                                                                 
=> {:inside_hash=>{:new_test=>"also_testing"}, :testing=>"test"}
[3] pry(main)> hash_one == hash_two                                                                                                                  
=> true
[4] pry(main)> hash_one.to_s == hash_two.to_s                                                                                                        
=> false
```

jmherbst:
I think my example might have been misleading.  Sorry about that. 

The issue is that the Sensu `event['action']` is a `Symbol` and the filter JSON's action is a `String`.

Need to find a way to make sure all values in the Sensu `event['action']` are of Class `String` before doing a comparison.

Ex with symbol:
```
> hash_one = {:testing=>"test", :inside_hash=>{:new_test=>:also_testing}}
 => {:testing=>"test", :inside_hash=>{:new_test=>:also_testing}}
> hash_two = {:inside_hash=>{:new_test=>"also_testing"}, :testing=>"test"}
 => {:inside_hash=>{:new_test=>"also_testing"}, :testing=>"test"}
> hash_one == hash_two
 => false
```

Is filtering by an `action` working at all? Or am I just mis-using it?

portertech:
I've thrown together https://github.com/sensu/sensu/pull/858, going to cut a beta release w/ the change.

portertech:
What version of Ruby are you using?

portertech:
Object comparison should return true.

jmherbst:
```
/opt/sensu/embedded/bin/ruby --version
ruby 2.0.0p353 (2013-11-22 revision 43784) [x86_64-linux]
```

portertech:
Master has string comparison of filter attribute value and event attribute value. The specs include a test for event `:action` :+1:

PR #854, opened 2015-01-29
Use the basic JSON gem to detect parsing errors in received data.

janedotx:
I discovered that if MultiJson is using Yajl to process JSON, and your Sensu messages are too big to fit in one packet, Sensu will not wait for all the packets to arrive before attempting to process the message. This PR fixes that.

### DIFF
--- a/lib/sensu/constants.rb
+++ b/lib/sensu/constants.rb
@@ -1,6 +1,6 @@
 module Sensu
   unless defined?(Sensu::VERSION)
-    VERSION = '0.16.0'
+    VERSION = '0.16.1'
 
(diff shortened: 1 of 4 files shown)


janedotx:
@portertech Do you think this PR is unnecessary? I was wondering if this was ever going to get merged in, since it's been sitting for a while.

portertech:
@janedotx thanks for the PR, but I think the whole socket protocol/stream implementation needs a rewrite, as messages are not guaranteed to be serialized.

portertech:
I'm going to close this PR as it cannot cleanly merge and the implementation will be getting a hammer thrown at it.

Issue #864, opened 2015-02-17
Dynamic parameters on check execution  


javrasya:
We have over thousand nodes(router, switches) to check them. Unfortunately, we can't use them as a sensu client. Somehow, there should be a common node which access all those routers and switch. My requirement is to execute a check on common node with different parameters. We are executing check as automated. Programatically, it is defined which node(int routers, switchs) will be used in checks. Their IP addresses for example. 

Somehow, check commands should be parameterised.  I saw the topic in the documentation which is named **Check command token substitution**. But it is only templating from client info. It should also be tokenising from a value coming from API. API can be like;

```json
{
     "check":"my_check",
     "subscribers":["test","web"],
     "kwargs":{
           "my_first_parameter":"value1",
           "my_second_parameter":"value2",
      }
}
```

warmfusion:
How would you identify the difference between a failure of my_check and my_check with different parameters. Sensu makes the assumption that checks are unique based on their host and check name, run time parametrisation changes that behaviour considerably. 

It may be more viable to use a cm to like puppet or chef to create standalone checks on your client and simply 'unroll' the parameters. 

Another option would be to just run the client, and stream your check results into its TCP listener.

portertech:
The upcoming features in 0.18 may help you https://github.com/sensu/sensu/blob/master/CHANGELOG.md#features Build out check definitions using check `source` to create JIT clients in the registry for external resources.

calebhailey:
Closing due to inactivity, although I'm noting that this may benefit from proposed changes in #935. Please feel free to create a new issue if this matter is not resolved. #monitoringlove

PR #865, opened 2015-02-17
Dynamic parameters on check execution from API

javrasya:
This is related with the issue #864 . Requirement description is defined there.

By the changes, it is support to send parameters on check execution. Here is its example of use;

This is check config;

``` json
{
  "checks": {
    "cron_check": {
      "handlers": ["default"],
      "command": "/etc/sensu/plugins/check-procs.rb -u :::user:::",
      "subscribers": [ "test" ],
     "interval": 60
    }
  }
}
```

It will result process check running under a user. Without my changes, to make it work, you should define `user` variable in your `client config`. But, what if you want to send some parameter to API with `/request` path on check execution. You couldn't do it.

Now, it can be done from API like;

``` json
{
    "check": "cron_check",
    "subscribers": ["test"],
    "kwargs":{
          "user":"ahmetdal"
     }
}
```

The process check script will now work with `ahmetdal` user. What if there is another variable with same name in client config. There is merging operation for `kwargs` coming from API and `client config`. `kwargs` prior than client config and it will override if same thing exists in `client config`

### DIFF
--- a/lib/sensu/api.rb
+++ b/lib/sensu/api.rb
@@ -416,6 +416,7 @@ module Sensu
         :subscribers => {:type => Array, :nil_ok => true},
         :kwargs =>  {:type => Hash, :nil_ok => true}
       }
+      
       read_data(rules) do |data|
         if settings.checks[data[:check]]
           check = settings.checks[data[:check]]
--- a/lib/sensu/api.rb
+++ b/lib/sensu/api.rb
@@ -413,7 +413,8 @@ module Sensu
     apost '/request/?' do
       rules = {
         :check => {:type => String, :nil_ok => false},
-        :subscribers => {:type => Array, :nil_ok => true}
+        :subscribers => {:type => Array, :nil_ok => true},
+        :kwargs =>  {:type => Hash, :nil_ok => true}
       }
       read_data(rules) do |data|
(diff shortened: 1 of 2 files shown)


javrasya:
can somebody help me, why i am getting the error for ruby version 1.9.3 while others are success;

```
  1) Sensu::Server can consume results
     Failure/Error: expect(MultiJson.load(latest_event_file)).to eq(event)
     MultiJson::ParseError:
       784: unexpected token at 'te"}'
     # ./spec/server_spec.rb:422:in `block (9 levels) in <top (required)>'
     # ./spec/helpers.rb:56:in `call'
     # ./spec/helpers.rb:56:in `block in timer'
     # ./spec/helpers.rb:62:in `async_wrapper'
     # ./spec/server_spec.rb:403:in `block (2 levels) in <top (required)>'
```

jonathanio:
I saw the same issue on a test build of mine for 1.9.2: There seams to be an intermittent error in the comparison somewhere.

javrasya:
Yes, you are probably right. After I triggered it to run one more time, tests are passed.

dcrawkstar:
+1

portertech:
We're not going to merge/implement this functionality, as we consider the security risk to be too great.

PR #873, opened 2015-02-25
Implement floating check mechanism

withnale:
*NB: Whilst this is functional it is not yet complete but has been shared for discussion. Further work on exception handling and specs will be required but wanted to get some feedback first*
***

At present (excluding standalone checks) sensu works based on a pub/sub subscription model for checks where clients that are subscribed to a given queue will all run the associated check.

### Floating Check
This PR defines a floating check which works somewhat differently and is used when you want a check to be run once only across a set of servers.

This sort of check would generally be used when the check has no direct actual tie to a single server. This might be for a number of reasons including...

1. The check is querying an external API such as AWS
2. The check is run on behalf of a piece of hardware that is unable to run the check, such as a router/load balancer.
3. The check is an aggregated check that generates results across a set of servers, such as a Graphite check utilising wildcards.
4. The check is exercising 'business logic' such as validating user experience.

### Configuration

At present, to that a given server wishes to be a member of a floating subscription group it must specify this in the client.json:

```json
 "client": {
    "name": "test_client",
    "address": "127.0.0.1",
    "subscriptions": [
      "test"
    ],
    "floating": [
      "floating_subscription"
    ]
 }
```

For a check to make use of this, it simply has to specify the transport to be direct (this defaults to 'fanout' which is the default value used when setting up a transport at present).

```json
"checks": {
    "floating_check": {
      "command": "my_sensu_check.rb",
      "subscribers": [
        "floating_subscription"
      ],
      "transport": "direct"
    }
```

### Implementation

When a "floating" subscription is associated with a client, that client will then listen for checks on the floating subscription AMQP queue. This is setup using the 'direct' transport rather than the 'fanout' transport used for normal subscriptions. Checks results are returned in the usual manner.

We need to avoid the situation where there is an attempt to setup a floating and normal subscription of the same name, and the associated errors when trying to register two differently configured transports with the same name. At present, to avoid this, a floating subscription is prefixed with the transport (i.e. direct_ 
). Ideally, I would have preferred to make this the default for normal subscriptions (i.e. fanout_) but was worried that this would cause regression problems.

Also, when the server receives results it needs to associate these with a client but we need to have a single client to represent the check rather than a client that will likely change between each iteration of the check. At present I have hardcoded this to a client called 'floating' as an initial implementation - am happy to hear better suggestions here.The implementation I have made is really a proof of concept and has been deliberately written to minimise the diff and the impact of the change, and try to ensure it works across multiple sensu-clients and for various API calls, and API consumers.

### DIFF
--- a/lib/sensu/client/process.rb
+++ b/lib/sensu/client/process.rb
@@ -230,12 +230,15 @@ module Sensu
       # transport pipe, named after the client subscription. The Sensu
       # client will receive JSON serialized check requests from its
       # funnel, that get parsed and processed.
-      def setup_subscriptions
-        @logger.debug("subscribing to client subscriptions")
-        @settings[:client][:subscriptions].each do |subscription|
-          @logger.debug("subscribing to a subscription", :subscription => subscription)
+      def setup_subscriptions_type(list, transport)
+        list.each do |subscription|
+          # Temporary fix - if transport is direct then prefix direct on the front of the subscription
+          # to avoid collisions but allow backward compat with earlier sensu clients
+          subscription = "direct_#{subscription}" if transport == :direct
+          @logger.debug("subscribing to a subscription",
+            :transport => transport, :subscription => subscription)
           funnel = [@settings[:client][:name], VERSION, Time.now.to_i].join("-")
-          @transport.subscribe(:fanout, subscription, funnel) do |message_info, message|
+          @transport.subscribe(transport, subscription, funnel) do |message_info, message|
             begin
               check = MultiJson.load(message)
               @logger.info("received check request", :check => check)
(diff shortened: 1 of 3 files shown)


bkett:
@withnale By switching from fanout to direct exchange type, what gains (performance or otherwise) are you expecting? My understanding is sensu creates a queue for each subscription. Fanout exchanges are very fast and generally you're only going to see performance reductions when publishing to a lot (thousands) of queues. It would seem that for these 'headless' checks (ala AWS queries for instance), I would think it is unlikely that there will be more than one or two subscriptions.

withnale:
Fixed the breaking checks.

withnale:
Bump - rebased after refactor.

portertech:
@withnale this functionality is something that I greatly desire. I have created a trello card here https://trello.com/c/5GN1GyXs/6-as-an-operator-i-would-like-to-have-a-check-request-go-to-a-single-subscriber-of-a-subscription The hardest part is "naming things", either the subscription name or the configuration attribute that indicates a direct "pipe" (direct exchange for RabbitMQ).

portertech:
We must remember that the Sensu transport is abstracted, certain naming conventions may not work when applied to different transports.

withnale:
I'll close this now. The roundrobin: prefix seems to do the same thing now. I think there are still some edge cases that need to be covered but will raise them as issues or PRs.

PR #875, opened 2015-02-27
allow reconnect on failure if required

davent:
When using AWS ElastiCache as a HA redis alternative sensu-server will shutdown when a read-slave is promoted as an error is received which causes `stop` to be called. This patch allows, if required, for the redis connection to closed so it can cleanly reconnect to the newly promoted redis master

### DIFF
--- a/lib/sensu/daemon.rb
+++ b/lib/sensu/daemon.rb
@@ -158,7 +158,11 @@ module Sensu
         @logger.fatal('redis connection error', {
           :error => error.to_s
         })
-        stop
+        if @settings[:redis][:reconnect_on_error]
+          @redis.close
+        else
+          stop
+        end
       end
       @redis.before_reconnect do
         unless testing?


johntdyer:
:+1:

portertech:
@davent did this trigger a reconnect?

davent:
@portertech It did yes. I have tested this against a 3 node AWS Redis ElastiCache Cluster. Failing over the nodes without this patch causes sensu-server to shutdown, but with it I was able to promote and de-promote nodes at random without any disruption to Sensu

portertech:
My tests prove otherwise :( The code suggests why https://github.com/portertech/em-redis/blob/unified/lib/em-redis/redis_protocol.rb#L481 I'll tinker with the Redis lib, see if we can get a `reconnect!()` method in there to trigger what we require.

portertech:
Here's the proposed em-redis patch https://github.com/portertech/em-redis/pull/3

portertech:
https://github.com/sensu/sensu/pull/891

portertech:
This Redis library (em-redis) now supports `reconnect_on_error`. This PR is no longer necessary, as the version of em-redis on Sensu core master has been bumped. :+1:

davent:
This does not appear to work. I will raise an Issue with my findings

PR #885, opened 2015-03-10
Fix occurrences count & reference original event.

jonathanio:
Change handling of event occurrences so that the count is only reset on an OK event: If an event is flapping between states and no flapping thresholds have been set its possible for it to never pass the occurrences threshold for the event to be handled. This will now continually increment occurrences so long as the check is reporting events without an OK status.

Additionally, added `:first_id` to the event data which will always contain the `:id` of the first event in this series of events, and kept so long as the events are not reporting OK or check is in a flapping state. This can be useful to help reference a batch of events, for example by using the `:id` and `:first_id` in the `Message-ID` or `In-Reply-To` headers of e-mails.

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -311,8 +311,13 @@ module Sensu
             :occurrences => 1
           }
           if check[:status] != 0 || flapping
-            if stored_event && check[:status] == stored_event[:check][:status]
+            if stored_event && check[:status] != 0
               event[:occurrences] = stored_event[:occurrences] + 1
+            end
+            if stored_event
+              event[:first_id] = stored_event.fetch(:first_id, event[:id])
+            else
+              event[:first_id] = event[:id]
             end
             event[:action] = flapping ? :flapping : :create
             @redis.hset("events:#{client[:name]}", check[:name], MultiJson.dump(event)) do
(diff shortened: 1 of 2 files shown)


portertech:
The event occurrence count is used to indicate re-occurrences of the same observed state (same check status), not how many times in a row the event is in a non-zero state. The event ID is only intended to provide a unique ID for the event occurrence, not a object reference. I am not keen on changing what the event occurrence count represents as it could break tooling/logic for many users. The event ID and it's use should certainly be revisited, as it's not offering users much at this time. Perhaps referencing the first event occurrence ID in the following recurrences would offer users something? If you're concerned about checks flapping between unhealthy states, perhaps the check history array should be used for filtering in your case?

petere:
If a check is set with occurrences = 5 and is reporting

    ..., 0, 0, 1, 1, 1, 1, 2, 2, ...

then, intuitively, the warning handlers should be run after the first "2", because the check has been warning or worse 5 times.  If we treat each state independently, then it could take up to 9 iterations (or worse, if flapping) before any handler is run, even though the user clearly meant 5.

Sensu tends to treat the check states independently rather than as a progression (e.g., you can set handlers for each state independently, occurrences are counted separately), but I think in practice users want the "or worse" behavior.

portertech:
Event IDs are now persistent, until the event is resolved (OK) :+1: The event occurrence count still represents the number of times the events has occurred with the same check exit status (this will never change). Events now include the `last_ok` and `last_state_change` timestamp attributes, which are better suited for notification filtering :+1:

PR #887, opened 2015-03-10
Use EventMachine 1.0.3 to fix Windows issues

portertech:


southwolf:
But em 1.0.3 is not compatible with Ruby 2.2, maybe just specify 1.0.3 for Windows only?

PR #888, opened 2015-03-11
Enable reconnect on failure for Transport.

jonathanio:
If the `server-server` service restarts, the channel is closed in RabbitMQ causing a failure in `sensu-api` which calls `stop`. This will (if configured) allow `sensu-api` to attempt a reconnection and continue rather than simply fail and abort (which will in turn affect event handling and Uchiwa).


portertech:
thelawapproves

PR #899, opened 2015-03-18
Fix for incorrect key to lookup :reconnect_on_error option.

jonathanio:
I put the incorrect key in for the transport name. This setting will allow the value to be included with the rest of the configuration for the transport (i.e. within rabbitmq's settings). Found when trying to expand Puppet configuration to support.

### DIFF
--- a/lib/sensu/daemon.rb
+++ b/lib/sensu/daemon.rb
@@ -184,7 +184,7 @@ module Sensu
       @transport = Transport.connect(transport_name, transport_settings)
       @transport.on_error do |error|
         @logger.fatal("transport connection error", :error => error.to_s)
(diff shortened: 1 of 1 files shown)


portertech:
This setting is intended to be within the transport context, as it is not specific to the Sensu transport in use.

pmyjavec:
May I ask why this is help up please?

portertech:
As of Sensu 0.18 https://github.com/sensu/sensu/blob/master/CHANGELOG.md#0180---2015-05-05 Sensu transport `reconnect_on_error` defaults to true, which applies to the loaded Sensu transport (default is RabbitMQ).

PR #889, opened 2015-03-12
Parameterized check extensions

rmc3:
This adds support for checks which have no 'command' key, but have an 'extension' key. Such check requests will run the extension whose name is the value of the 'extension' key. The check request hash will be passed through to the run() method of the extension.

This is intended to fix https://github.com/sensu/sensu/issues/866. See the comments on that issue for the rationale behind these changes, and the pattern of check extensions for which this is intended to add support.

This PR requires an changes to sensu-settings, which will be forthcoming in another PR.

One note: A local ```rake spec``` run showed many failures. This could just be an error with my build environment -- I see those failures when running it against ```master``` as well. We'll see if the tests pass in Travis :)

PR #902, opened 2015-03-19
add extensiona and opts to check payload

Igorshp:
Patch to go with https://github.com/sensu/sensu/pull/889

The server only sends check name and issued time to the clients.
This PR adds extension name and extension_opts if they exist.

@rmc3 I've used the format from your check example. Have you already solved this exact problem or are you using the checks in standalone mode?

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -412,7 +412,6 @@ module Sensu
         }
         payload[:command] = check[:command] if check.has_key?(:command)
         payload[:extension] = check[:extension] if check.has_key?(:extension)
-        payload[:extension_opts] = check[:extension_opts] if check.has_key?(:extension_opts)
 
         @logger.info("publishing check request", {
           :payload => payload,
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -412,7 +412,6 @@ module Sensu
         }
         payload[:command] = check[:command] if check.has_key?(:command)
         payload[:extension] = check[:extension] if check.has_key?(:extension)
-
         @logger.info("publishing check request", {
           :payload => payload,
           :subscribers => check[:subscribers]
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -411,6 +411,9 @@ module Sensu
           :issued => Time.now.to_i
         }
         payload[:command] = check[:command] if check.has_key?(:command)
+        payload[:extension] = check[:extension] if check.has_key?(:extension)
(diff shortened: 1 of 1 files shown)


Igorshp:
I've been porting the check_haproxy community plugin to an extension and that's when this came up.

here's the how I'm currently using it:
https://gist.github.com/Igorshp/8646d5d933b98f2e40cb#file-check_haproxy_extension-L126

looking for comments/suggestions.

Igorshp:
Starting to have a feeling that i'm supposed to use the name of the check to look up config in @settings[:checks]

Igorshp:
Ok, I have fully working haproxy extension ready that requires a check to have just name and extension.

I still feel that adding extension_opts would simplify things a bit, but on the other hand, doing a @settings[:checks] lookup does return more data. Might be useful.

rmc3:
:+1: I hadn't considered this when working on #889, I've only ever run Sensu as standalone with safe mode on. It does seem like we'd need to pass through check[:extension] for the flow in ```Sensu::Client::Process#process_check_request``` to work. Additionally, allowing check[:extension_opts] to be passed in the check request payload as @Igorshp suggests could allow parameters to be passed to extension runs when they're not running in safe mode.

@portertech Thoughts? Also, why not just pass the entire check definition object from the server to the client when issuing a check run request?

lib/sensu/server/process.rb
```diff
@@ -411,6 +411,8 @@ def publish_check_request(check)
           :issued => Time.now.to_i
         }
         payload[:command] = check[:command] if check.has_key?(:command)
+        payload[:extension] = check[:extension] if check.has_key?(:extension)
+
```
portertech: Please drop the extra whitespace/line :+1:

portertech:
For the meantime, you can add a local check definition containing the extension name, it will merge w/ the check requests.

Igorshp:
@portertech updated

portertech:
@Igorshp check definition merging w/ the check request payload is currently only done when the request contains a check command.

Igorshp:
@portertech 
Ah, found it now, I see what you mean.
https://github.com/sensu/sensu/blob/master/lib/sensu/client/process.rb#L214

Wouldn't it be good to do that globally for every check? That would remove the need to do the merging in every extension and , as you said,  remove the need for this PR.
Personally I prefer the check merge approach as it simplifies extension code a bit.

I've prepared PR.
https://github.com/sensu/sensu/pull/906

Igorshp:
Closing in favor of https://github.com/sensu/sensu/pull/906

PR #905, opened 2015-03-23
Fix eventmachine dependency for both Windows and other platforms

southwolf:
Fix #896 and #887 for good. Both sides should be happy :smile:

### DIFF
--- a/sensu.gemspec
+++ b/sensu.gemspec
@@ -16,7 +16,11 @@ Gem::Specification.new do |s|
   s.add_dependency "json" if RUBY_VERSION < "1.9"
   s.add_dependency "multi_json", "1.11.0"
   s.add_dependency "uuidtools", "2.1.4"
-  s.add_dependency "eventmachine", "1.0.3"
+  if RUBY_PLATFORM =~ /mingw/
+    s.add_dependency "eventmachine", "1.0.3" 
+  else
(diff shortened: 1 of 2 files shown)


southwolf:
Seems that Travis CI fails randomly. It works now.

kreynolds:
+1

sensu.gemspec
```diff
@@ -16,7 +16,11 @@ Gem::Specification.new do |s|
   s.add_dependency "json" if RUBY_VERSION < "1.9"
   s.add_dependency "multi_json", "1.11.0"
   s.add_dependency "uuidtools", "2.1.4"
-  s.add_dependency "eventmachine", "1.0.3"
+  if RUBY_PLATFORM =~ /mingw/
```
portertech: This cannot work when publishing the gem to rubygems, as it will use the platform it is pushed from.

portertech:
The Ruby platform is evaluated when the gem is built and published to Rubygems, not on install etc :( Going to close this in favour of https://github.com/sensu/sensu/pull/914 which may be replaced with a https://github.com/sensu/sensu-build change to install the latest eventmachine using devkit.

sodabrew:
The latest eventmachine definitely works on Windows; if you're seeing any trouble please report it?

PR #910, opened 2015-03-29
Implement check staleness checking

roobert:
I need to add tests but I thought I'd open a PR to get some feedback on the implementation.

This can be tested by adding an alert with @solarkennedy's excellent [sensu-shell-helper](https://github.com/solarkennedy/sensu-shell-helper), e.g:

```
sensu-shell-helper -j "\"executed\":\"$(date +%s)\"," -- echo check_event_keepalive
```

I've removed `execution:<client>:<check>` since `result` data is now stored in `result:<client>:<check>`, the API has been updated to reflect this change, too.

`check[:output]` has been limited to 256 bytes in length to prevent huge amounts of data being stored in redis - as suggested by @portertech.

To disable check keepalives for a check you can set the `:critical` and `:warning` thresholds to 0.

### DIFF
--- a/lib/sensu/api/process.rb
+++ b/lib/sensu/api/process.rb
@@ -170,7 +170,9 @@ module Sensu
           begin
             data = MultiJson.load(env["rack.input"].read)
             valid = rules.all? do |key, rule|
-              data[key].is_a?(rule[:type]) || (rule[:nil_ok] && data[key].nil?)
+              value = data[key]
+              (value.is_a?(rule[:type]) || (rule[:nil_ok] && value.nil?)) &&
+                rule[:regex].nil? || (rule[:regex] && (value =~ rule[:regex]) == 0)
             end
             if valid
               callback.call(data)
@@ -302,6 +304,23 @@ module Sensu
         end
       end
 
+      apost "/clients/?" do
+        rules = {
+          :name => {:type => String, :nil_ok => false, :regex => /^[\w\.-]+$/},
+          :address => {:type => String, :nil_ok => false},
+          :subscriptions => {:type => Array, :nil_ok => false}
+        }
+        read_data(rules) do |data|
+          data[:keepalives] = false
+          data[:timestamp] = Time.now.to_i
+          settings.redis.set("client:#{data[:name]}", MultiJson.dump(data)) do
+            settings.redis.sadd("clients", data[:name]) do
+              created!(MultiJson.dump(:name => data[:name]))
+            end
+          end
+        end
+      end
+
       aget "/clients/?" do
         response = Array.new
         settings.redis.smembers("clients") do |clients|
@@ -333,7 +352,7 @@ module Sensu
 
       aget %r{/clients/([\w\.-]+)/history/?$} do |client_name|
         response = Array.new
-        settings.redis.smembers("history:#{client_name}") do |checks|
+        settings.redis.smembers("result:#{client_name}") do |checks|
           unless checks.empty?
             checks.each_with_index do |check_name, index|
               result_key = "#{client_name}:#{check_name}"
@@ -430,7 +449,7 @@ module Sensu
               :subscribers => subscribers
             })
             subscribers.uniq.each do |exchange_name|
-              settings.transport.publish(:fanout, exchange_name, MultiJson.dump(payload)) do |info|
+              settings.transport.publish(:fanout, exchange_name.to_s, MultiJson.dump(payload)) do |info|
                 if info[:error]
                   settings.logger.error("failed to publish check request", {
                     :exchange_name => exchange_name,
--- a/lib/sensu/api/process.rb
+++ b/lib/sensu/api/process.rb
@@ -40,7 +40,7 @@ module Sensu
           setup_logger(options)
           set :logger, @logger
           load_settings(options)
-          set :api, @settings[:api]
+          set :api, @settings[:api] || {}
           set :checks, @settings[:checks]
           set :all_checks, @settings.checks
           set :cors, @settings[:cors] || {
@@ -55,8 +55,13 @@ module Sensu
 
         def start_server
           Thin::Logging.silent = true
-          bind = @settings[:api][:bind] || "0.0.0.0"
-          @thin = Thin::Server.new(bind, @settings[:api][:port], self)
+          bind = settings.api[:bind] || "0.0.0.0"
+          port = settings.api[:port] || 4567
+          @logger.info("api listening", {
+            :bind => bind,
+            :port => port
+          })
+          @thin = Thin::Server.new(bind, port, self)
           @thin.start
         end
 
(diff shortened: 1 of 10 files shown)


lib/sensu/server/process.rb
```diff
@@ -363,6 +362,10 @@ def process_check_result(result)
                 end
               end
             end
+            @logger.debug("updating result data", :check=> check)
+            result_truncated_output = result.dup
+            result_truncated_output[:output] = result[:output][0..256]
+            @redis.set("result:#{client[:name]}:#{check[:name]}", MultiJson.dump(result_without_output))
```
solarkennedy: How will a user clear this result table if they are sure they want to *really* remove a check?

lib/sensu/server/process.rb
```diff
@@ -572,6 +575,58 @@ def determine_stale_clients
...
+        @redis.smembers("clients") do |clients|
+          clients.each do |client_name|
+            @redis.keys("result:" + client_name + ":*") do |results|
+              results.each do |result_name|
+                next if result_name.split(':')[2] == 'keepalive'
+                @redis.get(result_name) do |result_json|
+                  next if result_json.nil?
+                  result = MultiJson.load(result_json)
+                  check_defaults = {
+                    :thresholds => {
```
solarkennedy: Are you sure you want a default threshold for everything?
Shouldn't this be more of an opt-in kinda thing? Where *only* if a check has set the :thresholds then it should be checked for staleness?

roobert: I took the defaults from the keepalive staleness check. The reason I think it's important to be opt-out rather than opt-in is because if you change a check config and it's nolonger valid JSON then your check will remain OK and you will have no idea the check isn't being run unless you manually inspect check times.. this is a very bad situation to be in.

lib/sensu/server/process.rb
roobert: I've updated `resolve_event` which gets called on check removal.

lib/sensu/server/process.rb
solarkennedy: But one of the *best* things about sensu is I can use configuration management to add and remove checks at will? If this is automatically added to every check, then won't it mean it will require manual intervention (or additional tooling) to auto-resolve all the checks that I "comment out" of my CM tool?


Additionally, this is a dramatic change in "before" behavior, where checks get a "natural" host dependency. If I have a host with 100 standalone checks, and it goes down, I'll get 100 alerts unless I *also* add an explicit host keepalive dependency. This should at least be noted. (and something I personally object to, I vote that this be opt-in)

lib/sensu/server/process.rb
solarkennedy: I don't see that change. I see it on the client delete api handler?

lib/sensu/server/process.rb
```diff
@@ -572,6 +575,58 @@ def determine_stale_clients
...
+      # the latest event is equal to or greater than a threshold,
+      # the check `:output` is set to a descriptive message, and
+      # `:status` is set to the appropriate non-zero value, and an
+      # alert is sent. If a check has been sending events, then
+      # no alert is published.
+      def determine_stale_checks
+        @logger.info("determining stale checks")
+        @redis.smembers("clients") do |clients|
+          clients.each do |client_name|
+            @redis.keys("result:" + client_name + ":*") do |results|
```
portertech: No Redis KEYS please. Not an operation you want to be doing.

lib/sensu/server/process.rb
```diff
@@ -572,6 +575,58 @@ def determine_stale_clients
...
+              results.each do |result_name|
+                next if result_name.split(':')[2] == 'keepalive'
```
portertech: Is this an indication that we should be using a different Redis data structure?

lib/sensu/server/process.rb
```diff
@@ -572,6 +575,58 @@ def determine_stale_clients
...
+                @redis.get(result_name) do |result_json|
+                  next if result_json.nil?
+                  result = MultiJson.load(result_json)
+                  check_defaults = {
```
portertech: Maybe use a separate method to constructe the check/result hash? As keepalives now does?

lib/sensu/server/process.rb
portertech: Undecided on opt-out vs in. I aggree w/ @solarkennedy, this feature could cause more of a headache if opt-out. Maybe check `heatbeat: bool`, also have to think about adhoc check results via the client socket.

roobert: Firstly, checks need to be manually removed through the API if their client config is removed, so in that sense this is no different, and will be handled at the same time as a check is normally removed.

Secondly, I have no strong opinions about the default thresholds being 0 (and therefore disabled) or the same as client keepalives. My argument for the former would be as I said before: it's currently possible for broken checks to be listed as 'OK', which I think is worse than changing the behaviour.

Either way, the functionality is what's important to me :)

lib/sensu/api/process.rb
```diff
@@ -382,7 +383,7 @@ def resolve_event(event_json)
                   settings.redis.smembers("history:#{client_name}") do |checks|
                     checks.each do |check_name|
                       settings.redis.del("history:#{client_name}:#{check_name}")
-                      settings.redis.del("execution:#{client_name}:#{check_name}")
+                      settings.redis.del("result:#{client_name}:#{check_name}")
```
portertech: @solarkennedy this is the deletion @roobert mentioned, removing the latest check result.

lib/sensu/server/process.rb
roobert: Good point, I'll create an index.

lib/sensu/server/process.rb
roobert: I think using mapped_hmset would be better than writing JSON as a string to redis. I did it this way as this is how it's done in other parts of the app.

lib/sensu/server/process.rb
roobert: OK, will change this.

roobert:
* set default thresholds to 0 for now, this minimizes unexpected behavioural changes at least
* created `results` index
* separate method to construct check keepalive check

had second thoughts about using mapped_hmset - realised it's probably better to have fewer redis reads so left things in the style as they are elsewhere.

roobert:
@portertech: how would you feel about an additional API end point - /result (or something) to allow access to result data? This could make the 'checks' page in Uchiwa more useful as it could display all existing checks. Status-0 checks could also display the truncated `result[:output]` data, along with additional meta-data for each check.

roobert:
this now depends on #922

jaxxstorm:
I would really like to see this merged, is there anything I can do to help?

roobert:
Now that #922 has been merged this could be merged. Probably needs some tests writing for it but I didn't want to write any until @portertech gave the OK on the implementation.

portertech:
@roobert perhaps start a fresh branch, cut from master?

Igorshp:
Would love to see this happen.
Actually came here after started using the sensu-shell-helper as well :)

roobert:
Rebased against master before I go on holiday.

I don't have an easy way to test this right now unfortunately but if it doesn't outright work it shouldn't be simple to fix.

portertech:
@roobert alternative, using check TTL https://github.com/sensu/sensu/pull/953

PR #914, opened 2015-03-31
Change gemspec to allow newer eventmachines, fixes ruby 2.2

kreynolds:
Less invasive gemspec change to allow eventmachine >1.0.3 for ruby 2.2, but still allow eventmachine 1.0.3 for windows if required. Replaces #905

### DIFF
--- a/sensu.gemspec
+++ b/sensu.gemspec
@@ -16,7 +16,7 @@ Gem::Specification.new do |s|
   s.add_dependency "json" if RUBY_VERSION < "1.9"
   s.add_dependency "multi_json", "1.11.0"
   s.add_dependency "uuidtools", "2.1.4"
-  s.add_dependency "eventmachine", "1.0.3"
+  s.add_dependency "eventmachine", "~>1.0", ">=1.0.3"
   s.add_dependency "sensu-em", "2.4.1"
   s.add_dependency "sensu-logger", "1.0.0"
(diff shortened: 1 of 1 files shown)


portertech:
I would prefer to get the Windows binary builds working again and bump the pin for all platforms.

kreynolds:
As I understand it, the windows builds work now, but require eventmachine <=1.0.3, and ruby 2.2 requires eventmachine >=1.0.4. Seems like adding flexibility in the Gemfile to support more platforms is better than waiting for the eventmachine maintainers to update their gem for windows no?

ctheune:
I think this is also blocking installation on OS X machines. I'm currently bundling sensu for NixOS / nixpkgs and I have to explicitly override the requirement. Would love this to get merged and released.

sensu.gemspec
```diff
@@ -16,7 +16,7 @@ Gem::Specification.new do |s|
   s.add_dependency "json" if RUBY_VERSION < "1.9"
```
bf4: also, fwiw, this only affects the build gem depending on the ruby version used to build it, so it's basically meaningless unless you're building the gem on 1.8

portertech:
Master uses EM 1.0.8.

bf4:
Fyi, You haven't been a very welcoming maintainer.

Issue #766, opened 2014-08-03
Node tags for aggregations


omribahumi:
I'm currently in a research process for replacing our monitoring system.
I've been looking at the aggregations API of Sensu and I think it's lacking an important feature: tagging (or grouping, or whatever). I'll explain:

Consider the following scenario: We have 10 different roles in our cluster.We're performing some generic system checks on all of them: Load average, CPU, Disk, Memory, Ping, SSH.
I want to aggregate based on role and check, for example, I would like to be notified if more than 1 "web" node has a "ping" check failure, without creating a separated "ping" check for "web" nodes.
With the way things are, there's no avoiding creating multiple checks for the same check.

Any thoughts?

Seraf:
Hello, I :+1:  this issue, I have (for cost reason and not my choice) only one sensu server for two regions and 4 environments. If I want to aggregate checks on a node type, it will do it for every regions and env.
I know it can be tricked with one sensu per env or namespacing check's name but it should be a core feature to be able matching regexp in the client's name, or support tagging imho. I guess it could use the client's subscriptions as tag ?

Seraf:
I was thinking about that, could we use subscriptions as tags ?
I mean for example I have nodes on aws in the region eu-west-1 and us-east-1. I have a subscription for each region containing the name of the region.
It should be great to have an aggregate check for keepalive check with the subscription 'eu' and another for the subscription 'us'.

What do you think about that ?

portertech:
We may be able to expand check aggregates or add another primitive/endpoint if it makes sense. Let's continue the discussion around some code https://github.com/sensu/sensu/pull/915

calebhailey:
Tagged this in #1218 for renewed discussion. Adding the "feature" label here as well.

PR #915, opened 2015-04-02
Implement aggregate grouping

roobert:
My use case is that I have multiple instances of various apps deployed to a single machine and monitored by a single sensu-client.

It's not possible to have duplicate check definitions with the same name so I've include a unique identifier in each check name, e.g: `fooapp-id-check_for_process`. When I add `aggregate: true` to my metadata, this results in the checks being written to separate buckets and not being aggregated properly.

This change allows you to specify an array of virtual check names to `aggregate:` enabling you to group checks into different (and multiple) buckets.

This builds upon the current effect of the `aggregate:` parameter and won't break existing check configs.

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -404,11 +404,14 @@ module Sensu
       # composted of a check `:name`, an `:issued` timestamp, and a
       # check `:command` if available. The check request is published
       # to a transport pipe, for each of the check `:subscribers` in
-      # its definition, eg. "webserver". JSON serialization is used
-      # when publishing the check request payload to the transport
-      # pipes. Transport errors are logged.
+      # its definition, eg. "webserver". Check requests are not 
+      # published if subdued. JSON serialization is used when 
+      # publishing the check request payload to the transport pipes.
+      # Transport errors are logged.
+      #
       #
       # @param check [Hash] definition.
+      # @param issued [Integer] issued timestamp.
       def publish_check_request(check, issued = nil)
         issued = Time.now.to_i unless issued
         if check_request_subdued?(check)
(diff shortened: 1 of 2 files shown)


roobert:
I've changed the implementation slightly as my initial attempt didn't take into account how the request scheduler works.

If `:aggregate` is set to a `String` then the check is assumed to be part of the specified group. Checks in an aggregate group are batched together and issued with the same `:issued` timestamp. A check can only belong to one group.

Existing behaviour is unaffected.

Seraf:
Exactly what I need. I also suggested to use subscription as tag : https://github.com/sensu/sensu/issues/766
Thanks for your work, I hope it will be merged

roobert:
Waiting for feedback from @portertech on this :)

portertech:
As discussed w/ @roobert, the future of the aggregate is unclear. We may need to split the functionality into two separate primitives/endpoints, to keep things well defined/clear.

Seraf:
Hello, any news about this feature ?

Pryz:
Any new on this ? Thanks

RyanJarv:
Looking into how to stop notifications based on groups of several different checks. For example if ~50% of any group of checks (one group per node, one group per service, one group per rack and one per datacenter) is down no notifications would be sent for that group. Being able to easily specify multiple aggregates would make this simple to implement.

maxlock:
Would love to have this, please!

calebhailey:
A proposal for "aggregates 2.0" has been published as a comment on issue #1218. Barring any objections, this proposal will be implemented in Sensu 0.24 and will resolve this PR.

portertech:
The release/0.26 branch includes "Aggregates 2.0" and supports multiple aggregates per check :+1: Thank you all for your ideas and input! Closing this pull request.

PR #920, opened 2015-04-10
Make use of the Redis reconnect_on_error setting

davent:
The `setup_redis` function was not making use of the `reconnect_on_error` setting as the `setup_transport` function does.


portertech:
It is built into the Redis client.

davent:
Haven taken another look at this I do not believe that this is correct. The clients `on_error` callback function takes a ruby block as it's parameter. the `lib/sensu/deamon.rb` file passed a block however all this block does is call the `stop` function:

@redis.on_error do |error|
        @logger.fatal("redis connection error", :error => error.to_s)
        stop
end

PR #921, opened 2015-04-12
Implement store result data

roobert:
Split from #915 as requested by @portertech.


roobert:
This PR had unintended changes in.

PR #922, opened 2015-04-12
split out store-result-data patch from #915

roobert:


portertech:
@roobert Love it, just giving it one more look over :+1:

lib/sensu/server/process.rb
```diff
@@ -363,6 +362,11 @@ def process_check_result(result)
                 end
               end
             end
+            @logger.debug("updating result data", :check=> check)
```
portertech: This should really be done in `store_check_result()`.

lib/sensu/server/process.rb
```diff
@@ -354,7 +359,7 @@ def process_check_result(result)
               result[:check]
             end
             aggregate_check_result(result) if check[:aggregate]
-            store_check_result(client, check) do
+            store_check_result(client, check, result) do
```
portertech: the `result` arg should not be required, since the provided `client` and `check` should have all of the data, no?

portertech: Or  perhaps `result` should be the only arg? Will have to take a look at both methods.

portertech:
Going to merge this and make a few minor changes :+1:

portertech:
https://github.com/sensu/sensu/pull/925

PR #923, opened 2015-04-13
allow filters to be defined in a check

Igorshp:
option to set filters in the check.

Some checks require variable levels of handling, and at the moment, a new handler need to be created for each filter variation, for example:

pagerduty_notify_instantly
pagerduty_notify_after_5_failed_attempts
pagerduty_notify_after_10_failed_attempts

Say I have a single check that i want to only notify after 5 failed attempts, but all others must notify instantly when critical.

this results in having handlers like:

pagerduty_notify_instantly
pagerduty_notify_after_5_failed_attempts
pagerduty_notify_after_10_failed_attempts

it would be simpler to have 1 handler, and simply define the filter inside the specific check that needs it.

### DIFF
--- a/lib/sensu/server/filter.rb
+++ b/lib/sensu/server/filter.rb
@@ -257,8 +257,8 @@ module Sensu
       # @param event [Hash]
       # @param callback [Proc]
       def event_filtered?(handler, event, &callback)
-        if handler.has_key?(:filters) || handler.has_key?(:filter)
-          filter_list = Array(handler[:filters] || handler[:filter])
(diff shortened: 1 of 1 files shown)


Igorshp:
Could anyone clue me in on what's happening here, and I can finish writing tests for this PR:
https://github.com/sensu/sensu/blob/master/spec/server/filter_spec.rb#L156

Why all the nesting?

portertech:
We do not want to add "filter" or "filters" to check definitions, as Sensu uses handler definitions to define/configure the result and event processing pipeline. Multiple handler definitions can be used to the same effect, without introducing more ways to influence the pipeline, potentially causing confusion.

Issue #541, opened 2013-05-19
Drop the client info requirement for event data


portertech:
531

ripasapa:
+1 we are using a lot e.g. AWS services and source name should be able to override. Putting source name to check_name is not recommended. Sensu is a great tool and i hope this custom json event with "source name whatever you like to put " can be implemented.

samion:
+1

jkamper:
https://github.com/sensu/sensu/issues/541#issuecomment-67759128

+1

roobert:
+1

roobert:
I've expanded @solarkennedy's patch so that masqueraded hosts are also visible through the API, see this ticket for details: https://github.com/Yelp/sensu/issues/1

Copse:
+1
If this get's implemented we can get rid of Nagios for good!

jtyr:
+1

johntdyer:
:+1: * :100:

johntdyer:
@portertech  any movement on this ?

josqu4red:
Being able to run checks on behalf on other hosts is a critical feature to me too; Any news on this ?

nien4521:
We are an it company that mostly sell and configure monitoring solutions (mostly nagios) for large infrastructure, and we are not allowed or we can't ( ip cam,switch, router, ...)  to install any kind of "client" on 70%  of the devices we have to monitor, sometime we only have acces to  :
- SNMP
- HTTP
- JUST A PING
- SSH/Telnet/Only a socket

The ONLY reason that we dont (yet) sell/install sensu is that we can't see those device as "client" in uchiwa and the api but instead we have to run our check on a client and we have as result in uchiwa 
localhost.device1.check1
localhost.device2.check1
....

The day we see a device without a sensu-client as a client in uchiwa/api we switch immediately to sensu.(curently we use nagios)
How come that a monitoring system doesn't have (natively) this kind of feature ?
I believe that the day this kind of feature is enabled in sensu everyone will drop nagios for sensu

TinajaLabs:
I have to say...  I was interested in sensu when we were using Icinga
1.8.4. Now Icinga 2.x is a re-written release and kicks ass.  Open source,
and pretty amazing.  The fact that the developers of sensu are not willing
to respond to their users means they are not really in the game.

Icinga 1.8.4 was just a clean-up of Nagios.  Icinga 2 is a re-write and the
team has built a really powerful monitoring application.

[...]

PR #924, opened 2015-04-14
Enable external resource monitoring

portertech:
Possible solution for https://github.com/sensu/sensu/issues/541

https://trello.com/c/eowfiJQ8/21-as-an-operator-i-would-like-to-monitor-external-resources-that-do-not-have-sensu-client-they-require-their-own-check-execution-h

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -352,7 +352,7 @@ module Sensu
           :subscriptions => [],
           :keepalives => false
         }
-        update_client_registry(client, callback)
+        update_client_registry(client, &callback)
       end
 
       # Retrieve a client (data) from Redis if it exists. If a client
@@ -370,7 +370,7 @@ module Sensu
             client = MultiJson.load(client_json)
             callback.call(client)
           else
-            create_client(client_key, callback)
+            create_client(client_key, &callback)
           end
         end
       end
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -586,6 +586,7 @@ module Sensu
             @redis.get("client:#{client_name}") do |client_json|
               unless client_json.nil?
                 client = MultiJson.load(client_json)
+                next if client[:keepalives] == false
                 check = create_keepalive_check(client)
                 time_since_last_keepalive = Time.now.to_i - client[:timestamp]
                 check[:output] = "No keepalive sent from client for "
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -49,7 +49,7 @@ module Sensu
         @logger.debug("updating client registry", :client => client)
         @redis.set("client:#{client[:name]}", MultiJson.dump(client)) do
           @redis.sadd("clients", client[:name]) do
-            callback.call
+            callback.call(client)
           end
         end
       end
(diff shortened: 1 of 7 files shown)


portertech:
Not sure about the use of `:keepalives => false`, may need a better attribute name to disable keepalive monitoring.

portertech:
Ready for review/feedback :+1:

portertech:
horsehead

lib/sensu/server/process.rb
```diff
@@ -552,6 +586,7 @@ def determine_stale_clients
             @redis.get("client:#{client_name}") do |client_json|
               unless client_json.nil?
                 client = MultiJson.load(client_json)
+                next if client[:keepalives] == false
```
solarkennedy: This kinda hints that any JIT client created by an arbitrary check-source will not be auto-pruned (seems legit)
Works for me. This has the effect that every client ever will remain in the client list until manually deleted right? I'm ok with that if you are (not like it hurts anything)

solarkennedy:
This looks good to me! I believe it will solve Yelp/sensu#1 !

I'm glad that you are using "source" as the name for the external client name key, as it makes it backwards compatible with Yelp's fork, but it also means that you... couldn't think of a better name either? :)

If you get this in mainline for a release, I'll volunteer to write docs?

lib/sensu/server/process.rb
portertech: Correct, keepalive events will never be created for these clients, therefor, deregistration handlers will never be executed. Manually deleting a client will resolve current events and cleanup history etc \o/

portertech:
@solarkennedy \o/ check "source" is the best I can do :P need to update the validation to use the client name regex tho, so there will be that constraint https://github.com/sensu/sensu-settings/blob/master/lib/sensu/settings/validators/check.rb#L72

portertech:
@solarkennedy is Yelp using check sources with special characters and space? 

PR: https://github.com/sensu/sensu-settings/pull/8

calebhailey:
This makes me happy. 

Thanks for your quick feedback @solarkennedy! :+1:

mholtrlc:
I'm trying to understand how to make this work, I don't see much in terms of documentation of this feature.

portertech:
@mholtrlc https://sensuapp.org/docs/latest/checks#definition-attributes check "source" which works with https://sensuapp.org/docs/latest/clients#jit-clients

mholtrlc:
ok, I was on the right rack with the jit-clients but missed the source attribute.

this is my configuration and I don't see anything showing up in the dashboard

```
{
	"checks": {
		"source": {
			"source": "VC-SRV-SW-A",
			"command": "/bin/true",
			"subscribers": [
				"vc-srv-sw-a"
			],
			"interval": 1
		},
		"unpublished": {
			"command": "/bin/true",
			"publish": false,
			"subscribers": [
				"vc-srv-sw-a"
			]
  		},
	    "VC-SRV-SW-A_Ping": {
	      "type": "metric",
	      "command": "/etc/sensu/plugins/check-ping.rb -h 10.10.11.149 --report",
	      "subscribers": ["vc-srv-sw-a"],
	      "interval": 10
	      "source": "vc-srv-sw-a"
	    }

	}
}
```

portertech:
@mholtrlc you are missing a comma after `"interval": 10`, invalid JSON, it will be logged and ignored.

mholtrlc:
haha i've made that mistake more than once lately.  I fixed that but am still not seeing it in the dashboard but I see it in the logs.

````
{"timestamp":"2015-05-20T09:08:01.147803-0700","level":"info","message":"publishing check request","payload":{"name":"VC-SRV-SW-A_Ping","issued":1432138081,"command":"/etc/sensu/plugins/check-ping.rb -h 10.10.11.149 --report"},"subscribers":["vc-srv-sw-a"]}
{"timestamp":"2015-05-20T09:08:01.618088-0700","level":"info","message":"publishing check request","payload":{"name":"source","issued":1432138081,"command":"/bin/true"},"subscribers":["vc-srv-sw-a"]}
{"timestamp":"2015-05-20T09:08:02.644661-0700","level":"info","message":"publishing check request","payload":{"name":"source","issued":1432138082,"command":"/bin/true"},"subscribers":["vc-srv-sw-a"]}
````

mholtrlc:
i'm thinking I don't have the correct configuration.  Here's what i've got:

client.json
````
{
  "client": {
    "name": "VC-SRV-SW-A",
    "address": "unknown",
    "subscriptions": [],
    "keepalives": false
  }
}
check.json
````
{
  "checks": {
    "VC-SRV-SW-A_Ping": {
      "source": "VC-SRV-SW-A",
      "command": "/etc/sensu/plugins/check-ping.rb -h 10.10.11.149 --report",
      "subscribers": [],
      "interval": 10
    }
  }
}

PR #925, opened 2015-04-15
cleaned up check result storage

portertech:
- only store check component of check result, client name is redundant
- replace check history set with result set
- added last_result to API client check history for bonus

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -207,21 +207,19 @@ module Sensu
       #
       # @param client [Hash]
       # @param check [Hash]
-      # @param result [Hash]
       # @param callback [Proc] to call when the check result data has
       #   been stored (history, etc).
-      def store_check_result(client, check, result, &callback)
-        @logger.debug("updating result data", :check=> check)
-        result_truncated_output = result.dup
-        result_truncated_output[:check][:output] = check[:output][0..256]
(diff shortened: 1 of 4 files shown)


lib/sensu/server/process.rb
```diff
@@ -207,21 +207,19 @@ def aggregate_check_result(result)
...
-        @redis.sadd("history:#{client[:name]}", check[:name])
+      def store_check_result(client, check, &callback)
+        @logger.debug("storing check result", :check => check)
+        @redis.sadd("result:#{client[:name]}", check[:name])
         result_key = "#{client[:name]}:#{check[:name]}"
-        history_key = "history:#{result_key}"
-        @redis.rpush(history_key, check[:status]) do
-          @redis.ltrim(history_key, -21, -1)
-          callback.call
+        check_truncated = check.merge(:output => check[:output][0..256])
```
portertech: merge() will create a new object, effectively doing the same as dup() with assignment.

portertech:
```
[1] pry(main)> foo = {:output => "foo"}                                                                                                              
=> {:output=>"foo"}
[2] pry(main)> bar = foo.merge(:output => "bar")                                                                                                     
=> {:output=>"bar"}
[3] pry(main)> foo                                                                                                                                   
=> {:output=>"foo"}
[4] pry(main)> bar                                                                                                                                   
=> {:output=>"bar"}
```

PR #928, opened 2015-04-22
close/stop client sockets when stopping the process

portertech:


calebhailey:
@portertech: "bewm", as they say. :boom:

portertech:
Travis is under maintenance until 12.

robx:
I see, thanks, didn't realize you had your own fork. I'll try to port the patch over later. Upstream event-machine seems somewhat dead.

PR #950, opened 2015-05-19
Round-robin subscriptions

portertech:
This PR add support for round-robin client subscriptions, allowing a single client in a subscription to receive a check request (1:1 instead of 1:N).

Trello card: https://trello.com/c/5GN1GyXs/6-as-an-operator-i-would-like-to-have-a-check-request-go-to-a-single-subscriber-of-a-subscription


portertech:
Ready for review :+1:

PR #951, opened 2015-05-19
Add /results routes

amdprophet:


lib/sensu/api/process.rb
```diff
@@ -725,6 +725,29 @@ def resolve_event(event_json)
           end
         end
       end
+
+      aget "/results/?" do
+        response = Array.new
+        settings.redis.smembers("clients") do |clients|
+          unless clients.empty?
+            clients.each_with_index do |client_name, client_index|
+              settings.redis.smembers("result:#{client_name}") do |checks|
```
portertech: if checks in empty, response will never be returned.

portertech:
:+1:

PR #964, opened 2015-06-01
Fixed flapping event resolution

portertech:


amdprophet:
:+1:

lib/sensu/server/process.rb
```diff
@@ -279,9 +279,11 @@ def check_history(client, check, &callback)
       def check_flapping?(stored_event, check)
         if check.has_key?(:low_flap_threshold) && check.has_key?(:high_flap_threshold)
           was_flapping = stored_event && stored_event[:action] == "flapping"
-          check[:total_state_change] >= check[:high_flap_threshold] ||
-            (was_flapping && check[:total_state_change] <= check[:low_flap_threshold]) ||
-            was_flapping
+          if was_flapping
+            check[:total_state_change] > check[:low_flap_threshold]
```
kraeuschen: hi,

i'm not sure if this is the expected behavior. if i understood right, the total state change on flapping checks needs to lower then the low flap threshold, not bigger.

PR #968, opened 2015-06-04
Use an EM timer for server socket timeout

portertech:


portertech:
This socket timeout works with the CRuby & JRuby EventMachine reactors.

PR #971, opened 2015-06-10
Make a checkttl expiration event always be critical

solarkennedy:
At yelp we never send warnings to our more critical handlers (pagerduty).

I want to be able to allow people to set a ttl on ephemeral checks and give them at least the option to page, but if it is alway set to warning, that can never happen?

I don't believe in warnings. :(

If you don't agree with this change, would you accept a more complicated PR that... set multiple thresholds like the keepalive thresholds? (but I would rather KISS here and have it set to always be crit)

cc @hashbrowncipher @bobtfish

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -670,8 +670,8 @@ module Sensu
                     time_since_last_execution = Time.now.to_i - check[:executed]
                     if time_since_last_execution >= check[:ttl]
                       check[:output] = "Last check execution was "
-                      check[:output] << "#{time_since_last_execution} seconds ago"
-                      check[:status] = 1
+                      check[:output] << "#{time_since_last_execution} seconds ago. (Threshold: #{check[:ttl]} seconds)"
+                      check[:status] = 2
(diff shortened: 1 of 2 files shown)


solarkennedy:
@portertech can I get some feedback on this and #972? I'm very eager to use this new feature but I don't want have to through the pain of forking again to make it usable.

portertech:
I do not believe a check failing to execution in a timely manner (or at all) warrants a critical status. A check TTL event could be the result of a variety of causes, sometimes more than one. I too would prefer to keep it simple, a single numeric threshold, as it is now. You could use a Sensu filter to match events with a check status of 1 and the TTL check output: `eval: value.start_with?("Last check execution")`, and apply the filter to a special pagerduty handler.

solarkennedy:
I think it depends on the check, but I agree that the simplicity is good and that this specific policy can be applied in other ways.

solarkennedy:
But, I would be curious if there are others who share my opinion, so others please perk up if you also see this is a bug or have other alternatives.

PR #972, opened 2015-06-10
Make force-resolving set a null TTL so that you can actually get rid of checks with ttls somehow

solarkennedy:
The check ttl feature is the bomb.
But it needs *some way* to actually get rid of checks after you no longer want them.

The problem is: how do you let sensu know that you no longer want it to keep reminding you that it is failing? One way to do that is to set `ttl:null` in the manual resolve. (sensu-cli/uchiwa/etc)

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -666,7 +666,7 @@ module Sensu
                 @redis.get("result:#{result_key}") do |result_json|
                   unless result_json.nil?
                     check = MultiJson.load(result_json)
-                    next unless check[:ttl] && check[:executed]
+                    next unless check[:ttl] && check[:executed] && !check[:force_resolve]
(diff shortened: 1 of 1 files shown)


portertech:
Hmm, is this even needed?

portertech:
Ah yes, https://github.com/sensu/sensu/blob/master/lib/sensu/server/process.rb#L398

portertech:
Maybe we should add `!check[:force_resolve]` to https://github.com/sensu/sensu/blob/master/lib/sensu/server/process.rb#L669 instead?

portertech:
That way the client check history will still have the correct check TTL value.

solarkennedy:
I can do that

portertech:
@solarkennedy missing the `&&`

solarkennedy:
Er, whoops, and the tests reflect that. Updated.

portertech:
:+1:

PR #975, opened 2015-06-17
Bump multi_json to 1.11.1

darix:


darix:
When you bump multi_json maybe also accept and release https://github.com/sensu/sensu-spawn/pull/5 and bump that requires too.

PR #983, opened 2015-06-24
log with debug level instead of info in filter_event

somic:
When an event is filtered out and not sent to a handler, right now this gets logged at level=info. I think it could be better to log this with level=debug.

There are pros and cons to either approach, I was just wondering what people think.

### DIFF
--- a/lib/sensu/server/filter.rb
+++ b/lib/sensu/server/filter.rb
@@ -296,14 +296,14 @@ module Sensu
           "handler is subdued"
         end
         if filter_message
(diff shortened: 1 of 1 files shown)


portertech:
I do like to know that an event was filtered without having to enable debug logging, however, it does indeed get noisy at times, particularly in certain use cases (e.g. filtering metric events).

portertech:
We believe that Sensu should continue to use the info log level for event filtering.

PR #1015, opened 2015-08-04
[clients] resolve inconsistent state in redis

quodlibetor:
I'm not sure exactly what can cause this state to occur, but it's
possible for sensu to have a client name in the clients list that does
not have a paired `client:` config.

When this happens, the api server restarts/crashes every time you hit
`/clients/`. And unfortunately you can't DELETE the client because `adelete` calls the exact same json.parse functionality as list does.

I'm not sure what the best way to handle this is, this definitely resolved the issue for me, but it seems plausible that much of `adelete` should be factored out into its own function so that we can delete all of the metadata around a partially inconsistent client.

### DIFF
--- a/lib/sensu/api/process.rb
+++ b/lib/sensu/api/process.rb
@@ -337,14 +337,18 @@ module Sensu
                 if client_json.nil?
                   settings.logger.error("redis returned nil for #{client_name}")
                   settings.redis.srem("clients", client_name)
-                  next
+                else
+                  response << MultiJson.load(client_json)
                 end
-                response << MultiJson.load(client_json)
+                if index == clients.size - 1
+                  body MultiJson.dump(response)
+                end
               end
             end
+          else
+            body MultiJson.dump(response)
           end
         end
-        body MultiJson.dump(response)
       end
 
       aget %r{^/clients?/([\w\.-]+)/?$} do |client_name|
--- a/lib/sensu/api/process.rb
+++ b/lib/sensu/api/process.rb
@@ -334,16 +334,17 @@ module Sensu
           unless clients.empty?
             clients.each_with_index do |client_name, index|
               settings.redis.get("client:#{client_name}") do |client_json|
-                response << MultiJson.load(client_json)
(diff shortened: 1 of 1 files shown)


lib/sensu/api/process.rb
```diff
@@ -334,7 +334,17 @@ def resolve_event(event_json)
...
+                begin
+                  response << MultiJson.load(client_json)
+                rescue MultiJson::ParseError
+                  settings.logger.error(
+                    "issue loading json for client #{client_name}: " \
+                    "redis returned #{client_json.inspect}")
+                  if client_json.nil?
+                    settings.redis.srem("clients", client_name)
+                  end
+                  next
```
portertech: Wouldn't `next` break the async response in the event of an error in the last client?

lib/sensu/api/process.rb
```diff
@@ -334,7 +334,17 @@ def resolve_event(event_json)
           unless clients.empty?
             clients.each_with_index do |client_name, index|
               settings.redis.get("client:#{client_name}") do |client_json|
-                response << MultiJson.load(client_json)
+                begin
+                  response << MultiJson.load(client_json)
+                rescue MultiJson::ParseError
```
portertech: We could just have a `if client_json.nil?` conditional, instead of catching the parser errors.

quodlibetor: That's true. It seems unlikely that we would have a partial or invalid json object in there. Originally I had no idea what was happening so I just caught everything.

quodlibetor:
Travis failed only on ruby 1.9.3, I can't tell how likely that is to be caused by my change.

lib/sensu/api/process.rb
quodlibetor: That's a good point, fixed.

quodlibetor:
Now travis failed only on 1.8.7, but it was the same test that failed.

portertech:
I restarted the 1.8.7 build.

portertech:
Thank you :+1:

PR #1022, opened 2015-08-10
Improved leader election and resignation

### LINKED ISSUE #1019: [ SENSU Server ] : The dangers of multiple masters/leaders - My story
Hi, Sean. I hope you are well. There is something I wanted to share with you. I recently discovered that my infrastructure was occasionally running multiple masters. In fact I was DoS(sing) myself. 

After thorough investigations I learnt that blocking calls from extensions (I simulated artificial delays) and/or client traffic and extensive read and write activity to Redis (I had around 3000 stashes and growing number of clients, in the hundreds; plus Sensu is storing all check results since version “0.18.0” and it has a cost) were just enough to block the event loop and prevent the server from updating the “leader” lock. (i.e.: I have system traces for both cases) In consequence, I was ending up with 16+ masters and a near-death experience, operationally speaking.

I also noticed these freezes were sometimes taking minutes and generated gaps in alarm and metrics coverage. A big issue in other words.

Grateful if you can share your thoughts on the matter. There are multiple things to it. One would be to force the “frozen” master to resign once it has detected a switch in leadership. That’s something that I/my team may contribute back with a PR shortly. It should also include a way to record the current master/leader's name in Redis. 

The second idea would be to have a new server role on the top of the current leader and worker instances. That would be a “publisher” solely responsible for broadcasting check requests and perhaps something more. (i.e.: in fact that’s one of my big concerns besides the DoS)

I am afraid that at scale, thousands of clients and lots of checks with a mix of metric and regular, one master with more than enough responsibilities: updating stale items, writing check results, publishing back to Rabbit, sending check requests, etc… will not cope. That’s where I am coming from with my message. 

I hope it generates some good ideas and comments from you and the community. Happy to share more details if needed…

Have a nice day.

Regards, Andrei

P.S.: the above was tested with “0.18.x” and “0.19.x” on Oracle Linux.

portertech:
Fixes #1019

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -746,6 +746,51 @@ module Sensu
         setup_check_result_aggregation_pruner
       end
 
+      def create_lock_timestamp
+        (Time.now.to_f * 1000).to_i
+      end
+
+      def become_the_leader(timestamp)
+        unless @is_leader
+          @logger.info("i am now the leader")
+          @is_leader = true
+          @last_leader_lock_timestamp = timestamp
+          leader_duties
+        else
+          @logger.debug("i am already the leader")
+        end
+      end
+
+      # Resign as leader, if the current process is the Sensu server
+      # leader. This method cancels and clears the leader timers,
+      # those with references stored in the timers hash under
+      # `:leader`, and `@is_leader` is set to `false`.
+      def resign_as_leader
+        if @is_leader
+          @logger.warn("resigning as leader")
+          @timers[:leader].each do |timer|
+            timer.cancel
+          end
+          @timers[:leader].clear
+          @is_leader = false
+        else
+          @logger.debug("not currently leader")
+        end
+      end
+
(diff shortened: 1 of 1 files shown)


agoddard:
also :+1: on branch name

portertech:
The ID is pretty random https://github.com/sporkmonger/uuidtools/blob/master/lib/uuidtools.rb#L219-L224

agoddard:
getinsertpic.com

portertech:
```
127.0.0.1:6379> keys *
1) "leader"
2) "lock:leader"
127.0.0.1:6379> get leader
"c7626e8f-ee06-4db5-9536-4318fb329bd3"
127.0.0.1:6379> get lock:leader
"1439329137718"
```

portertech:
Idea of using a leader ID came from https://github.com/thomsonreuters/sensu-1/pull/1

agoddard:
getinsertpic.com

PR #1024, opened 2015-08-14
Adjust how check output truncating works.

roobert:
Truncate output for metric checks to one line followed by '...' on a new
line.

Normal check output isn't truncated as generally it is quite short
and in any case the output gives necessary context to the check.

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -216,10 +216,16 @@ module Sensu
       def store_check_result(client, check, &callback)
         @logger.debug("storing check result", :check => check)
         @redis.sadd("result:#{client[:name]}", check[:name])
-        result_key = "#{client[:name]}:#{check[:name]}"
-        if check[:type] == "metric" && check[:output].split("\n").length > 1
-          check = check.merge(:output => "#{check[:output].split("\n")[0]}\n...")
+        if check[:type] == 'metric'
+          output = check[:output].split("\n")
+          if output.first.length < 255 && output.length = 1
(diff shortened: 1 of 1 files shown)


lib/sensu/server/process.rb
```diff
@@ -213,8 +217,10 @@ def store_check_result(client, check, &callback)
         @logger.debug("storing check result", :check => check)
         @redis.sadd("result:#{client[:name]}", check[:name])
         result_key = "#{client[:name]}:#{check[:name]}"
-        check_truncated = check.merge(:output => check[:output][0..256])
-        @redis.set("result:#{result_key}", MultiJson.dump(check_truncated)) do
+        if check[:type] == "metric" && check[:output].split("\n").length > 1
```
portertech: Still possible for very large single line outputs.

lib/sensu/server/process.rb
```diff
@@ -212,9 +216,16 @@ def aggregate_check_result(client, check)
       def store_check_result(client, check, &callback)
         @logger.debug("storing check result", :check => check)
         @redis.sadd("result:#{client[:name]}", check[:name])
+        if check[:type] == 'metric'
+          output_lines = check[:output].split("\n")
+          output = output_lines.first
+          if output_lines.size > 1 || output.length > 255
+            output = output[0..255] + "\n..."
+          end
+          check = check.merge(:output => output)
```
portertech: We probably want to create a new object, like `check_truncated`, or we are mutating event data! 0_0

PR #1039, opened 2015-09-04
Locking via API, wildcard results fetching

keymone:
`GET /lock/<name>/<holder>/<ttl>` - give some coordination to standalone checks.

Standalone checks are isolated to sensu-client process and i don't [see an easy way / think it's good idea / feel there is any need] to expose sensu-server process state (is_leader) to them. Instead locking via simple get request could be simple and functional enough to coordinate these checks.

`GET /results/*/<check_name>` - allow fetching check results for all clients in sensu cluster.

To implement something like a cluster check at this point means either fetching list of all clients + N requests for a check result we need to observe, or expose redis connection details to sensu-client process and rely on sensu-implementation details of how check results are stored. I think it's useful/simple enough to just allow GET on results:*:<check_name>. Yes, that is still N+1 problem but at least it's within single API request.

### DIFF
--- a/lib/sensu/api/process.rb
+++ b/lib/sensu/api/process.rb
@@ -800,8 +800,12 @@ module Sensu
         settings.redis.keys(keys) do |keys|
           if keys && keys.size > 0
             EM::Iterator.new(keys).map(
-              proc {|key, iter| settings.redis.get(key) {|data| iter.return(data) } },
-              proc {|responses| body MultiJson.dump(responses) }
+              lambda do |key, iter|
+                settings.redis.get(key) do |data|
+                  iter.return(key.split(':')[1] => data)
+                end
+              end,
+              lambda { |res| body MultiJson.dump(res.inject({}, &:merge)) }
             )
           else
             not_found!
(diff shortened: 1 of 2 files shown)


keymone:
i don't see how that single failing example is relevant to my changes.

solarkennedy:
The wildcard grouping is cool.

We can do the lock stuff now as a server-side function with this:
https://sensuapp.org/docs/0.20/clients#round-robin-client-subscriptions (but it does require us to remember to add subscriptions to all the right places instead of using standalone checks)

portertech:
I kicked the build to get it to pass :+1: You may want to rebase on master to get some of the spec/test improvements.

lib/sensu/api/process.rb
```diff
@@ -794,6 +794,48 @@ def resolve_event(event_json)
           end
         end
       end
+
+      aget %r{^/results?/\*/([\w\.-]+)/?$} do |check_name|
+        keys = "result:*:#{check_name}"
+        settings.redis.keys(keys) do |keys|
+          EM::Iterator.new(keys).map(
```
portertech: `EM::Iterator` creates anonymous classes, causing memory issues on JRuby. We should probably use a combination of `Proc` and `EM::next_tick`, like: https://github.com/sensu/sensu/blob/master/lib/sensu/server/filter.rb#L262-L272

lib/sensu/api/process.rb
```diff
@@ -794,6 +794,48 @@ def resolve_event(event_json)
           end
         end
       end
+
+      aget %r{^/results?/\*/([\w\.-]+)/?$} do |check_name|
+        keys = "result:*:#{check_name}"
+        settings.redis.keys(keys) do |keys|
```
portertech: AFAIK, not suppose to use Redis KEYS in your application.

portertech:
@keymone @solarkennedy are you still using an API endpoint for locking, in regards to standalone checks?

lib/sensu/api/process.rb
amdprophet: That is correct. The KEYS command should only be used for debugging purposes. More information (including potential workarounds) at http://redis.io/commands/KEYS.

solarkennedy:
@portertech we are. Our more advanced standalone checks currently use redis for locking, which we are trying to get away from. Other teams that didn't know about sensu internals used zookeeper (because that is what they new)

portertech:
Closing this pull-request, as the Sensu API has since been rewritten to use EM HTTP Server. Sensu now supports round-robin subscriptions, enabling efficient and reliable cluster monitoring from 1 of N hosts etc.

PR #1042, opened 2015-09-06
Fixes a broken test case for validation in Client API

miyakawataku:
This pull request fixes a broken test case for validation in Client API, which is named "can not create a client with an invalid post body".

The original test case intends to prepare three invalid sets of parameters (``invalid_name``, ``missing_address``, and ``invalid_subscriptions``), and asserts that Client API responds 400 for all of them.  However it actually tests only one set of parameters, because it uses Hash#dup to copy parameters, and thus an internal hash (``options[:body]``) is shared for them. As a result, the test case passes even if one of the validation rules is removed.

This pull request divides the test case to three individual cases so that the parameter set is not shared.

### DIFF
--- a/spec/api/process_spec.rb
+++ b/spec/api/process_spec.rb
@@ -118,35 +118,53 @@ describe "Sensu::API::Process" do
     end
   end
 
-  it "can not create a client with an invalid post body" do
+  it "can not create a client with an invalid post body (invalid name)" do
     api_test do
       options = {
         :body => {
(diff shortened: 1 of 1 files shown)


portertech:
@miyakawataku Nice catch! :+1: I was unaware that a Hash dup() is not recursive.

portertech:
I will merge this and make further changes to the API specs if necessary. Thank you!

PR #1050, opened 2015-09-09
Fixed metric check output truncation when output is empty

portertech:
Turns out `split("\n")` on an empty string produces an empty array.
```
[1] pry(main)> check = {:output => "foo"}                                                                                                            
=> {:output=>"foo"}
[2] pry(main)> output_lines = check[:output].split("\n")                                                                                             
=> ["foo"]
[3] pry(main)> check = {:output => ""}                                                                                                               
=> {:output=>""}
[4] pry(main)> output_lines = check[:output].split("\n")                                                                                             
=> []
```

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -210,7 +210,7 @@ module Sensu
         case check[:type]
         when "metric"
           output_lines = check[:output].split("\n")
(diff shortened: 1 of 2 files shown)


agoddard:
ah snap! nice catch :+1:

PR #1056, opened 2015-09-21
Remove EM::Iterator from event filtering

portertech:
- EM::Iterator uses anonymous classes (not GC'd on JRuby)

- Using a Proc w/ EM::next_tick executes filters serially, returning immediately if a filter matches (filtered), performing better than em iterator.


portertech:
This is the reason https://github.com/eventmachine/eventmachine/blob/master/lib/em/iterator.rb#L163-L168

amdprophet:
:+1:

portertech:
hb591b49e

PR #1059, opened 2015-09-22
Improve the API to handle requests while state changes

portertech:


portertech:
Creating JIT clients in the registry with:

```
for i in {1..1000}; do echo "{\"name\": \"app_01\", \"output\": \"could not connect to mysql\", \"status\": 1, \"source\": \"$i\"}" | nc localhost 3030; done
```

Going to reproduce a known error.

lib/sensu/api/process.rb
```diff
@@ -335,18 +335,9 @@ def resolve_event(event_json)
             clients.each_with_index do |client_name, index|
               settings.redis.get("client:#{client_name}") do |client_json|
                 unless client_json.nil?
-                  begin
-                    response << MultiJson.load(client_json)
-                  rescue MultiJson::ParseError => error
```
portertech: Only observed JSON parsing errors are when `client_json` was `nil`, before we added the conditional. AFAIK Redis does not support partial writes and the Redis client library only calls the callback after a full read.

amdprophet:
:+1:

PR #1060, opened 2015-09-23
Fixes misuse of regexp metacharacters "^" and "$"

miyakawataku:
This pull request fixes misuse of regexp metacharacters "^" (start of line) and "$" (end of line) all over the source tree. Those metacharacters are wrongly used where "\A" (start of string) and "\z" (end of string) should be used instead.

Consequently, this pull request fixes bugs as follows.

- Client API has been accepting multiline client names
- API has been accepting multiline input for integral parameters
- Client Socket has been accepting multiline check names and source names
- "occurrence" scripts for filters has not been evaluated correctly if they contain "\neval:"

### DIFF
--- a/lib/sensu/api/process.rb
+++ b/lib/sensu/api/process.rb
@@ -190,7 +190,7 @@ module Sensu
         end
 
         def integer_parameter(parameter)
-          parameter =~ /^[0-9]+$/ ? parameter.to_i : nil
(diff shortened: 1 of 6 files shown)


portertech:
90cc80c0-7afb-11e4-9130-b5de5f5b2ebf

PR #1063, opened 2015-09-26
Fix randomly occurring false negative test results for truncation of check output

miyakawataku:
This pull request fixes a test for truncation of check output which causes false negative randomly.

The test tries to generate a string of 256 characters and asserts that the string is truncated to 255 characters. But actually it sometimes (once in 16 times exactly) generates less than 256 characters and causes a false negative test result. For example:

https://travis-ci.org/sensu/sensu/jobs/79523141

This pull request justifies the string to 256 characters.

I hope it will reduce big red cross marks on the PR list and let us be happier ;)


portertech:
Thank you :+1:

PR #1066, opened 2015-10-03
Reduce failures of the test case for check result consumption

miyakawataku:
The test case "can consume results" fails often because of race condition or timeout. This pull request reduces the chance of them.

## Failure mode 1: timeout

The test may fail because of timeout. Examples:

- https://travis-ci.org/sensu/sensu/jobs/81634354
- https://travis-ci.org/sensu/sensu/jobs/79522099

This pull request increases the timeout threshold.

## Failure mode 2: unexpected occurrences count

``occurrences`` attribute of the event may differ from the expectation. Examples:

- https://travis-ci.org/sensu/sensu/jobs/79522897
- https://travis-ci.org/sensu/sensu/jobs/78767727

Probably the failures occur because two check results published by the test case are processed concurrently and cause race condition. It cannot easily be avoided because Redis lacks an atomic read-modify-write feature.

This pull request add a wait between the publishments of check results. It increases the chance that the check results are processed serially.

## Failure mode 3: mismatch between Redis and the output of a bridge

The output by ``Sensu::Extension::LatestEventFile`` may differ from the content in Redis. Example:

- https://travis-ci.org/sensu/sensu/jobs/68232751

Probably this situation occurs when EventMachine processes a task for the latter check result first. I cannot decide if this situation is worth to be fixed.

This pull request only reduce the chance of the situation, in the same way as the failure mode 2.

### DIFF
--- a/spec/helpers.rb
+++ b/spec/helpers.rb
@@ -53,9 +53,9 @@ module Helpers
     end
   end
 
-  def async_wrapper(&callback)
+  def async_wrapper(timeout_secs = 15, &callback)
     EM::run do
-      timer(15) do
+      timer(timeout_secs) do
         raise "test timed out"
       end
       callback.call
--- a/spec/server/process_spec.rb
+++ b/spec/server/process_spec.rb
@@ -159,35 +159,37 @@ describe "Sensu::Server::Process" do
           redis.set("client:i-424242", MultiJson.dump(client)) do
             result = result_template
             transport.publish(:direct, "results", MultiJson.dump(result))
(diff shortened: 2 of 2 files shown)


portertech:
@miyakawataku Thank you for this PR :+1: This has been needed for a long time.

I believe failure modes #2 & #3 are due to result processing time, I am not convinced that they are related to concurrency etc, since the Redis client queues Redis commands (serial) and Redis is also single threaded.

I am happy with your solution.

hb591b49e

Issue #1067, opened 2015-10-05
Exit codes >1 hide RESOLVEs


jeremybz:
Noticed a strange behaviour with handlers: when handling checks which have multiple possible statuses,  handlers ignore RESOLVEs for certain sequences (for example: 1,1,1,2,0,0,0).

I isolated it to results of the following standalone check:
{  "checks": { "EXIT_CODE_TESTING": {
      "command": "perl -e 'exit `cat /tmp/exitcode.txt`'",
      "interval": 5,
      "occurrences": 3,
      "standalone": true,
      "handlers": [
        "default"
      ] } } }

Changing /tmp/exitcode.txt to 0, 1, 2 gives the following results:
0, 0, 0, 1, 1, 1, 0,  => Works as expected, all handlers report a FAIL and a RESOLVE
0, 0, 0, 1, 1, 1, 2, 2, 2, 0 => Works as expected, FAIL, OTHER, RESOLVE
0, 0, 0, 1, 1, 1, 2, 0, 0, 0, 0... => Only FAIL, no RESOLVE. Server log reports: "not enough occurrences" message in all handlers.
0, 0, 0, 1, 1, 1, 2, 2, 0, 0, 0, 0... => Same problem, no RESOLVE.

I think this is due to server's process.rb, which does not update event[:occurrences] if the current fail status is different from the previous fail status, but I don't know enough about sensu (or ruby) to actually figure it out and submit a patch.
https://github.com/sensu/sensu/blob/master/lib/sensu/server/process.rb#L345
I am using version 0.20.3

portertech:
@jeremybz event "occurrences" represents a count of recurrences of a single state (e.g. status: 2), not non OK as you may be expecting. The filtering logic that you are referring to is in the sensu-plugin library, that is used by most community event handlers https://github.com/sensu-plugins/sensu-plugin/blob/master/lib/sensu-handler.rb#L97

portertech:
@jeremybz we could use event check history (21 exist status codes) for non OK recurrence filtering, instead of event occurrences.

jeremybz:
I tried adding a new 'nonzero_occurrences' variable to handle the problem, as parsing the history left me feeling uneasy.  I am not skilled at git or ruby, please let me know if this is horribly wrong.

https://github.com/sensu-plugins/sensu-plugin/pull/97
https://github.com/sensu/sensu/pull/1069

jeremybz:
[...]

PR #1069, opened 2015-10-07
Suggested fix for issue 1067: add 'nonzero_occurrences' to solve igno…

jeremybz:
…red Resolves for status history like: 1,1,1,2,0. https://github.com/sensu/sensu/issues/1067
Requires a PR for sensu-plugin as well (same title)

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -338,9 +338,13 @@ module Sensu
             :client => client,
             :check => check,
             :occurrences => 1,
+            :nonzero_occurrences => 1,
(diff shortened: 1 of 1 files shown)


portertech:
Or should we store the timestamp for the last OK status? Then no time calculation w/ counts and interval are required. What do you think? 'last_ok'?

jeremybz:
Ok, my "patches" are awful, they only make stuff worse.  What about an "alert_triggered" boolean, which gets set once occurrences is reached, and only unset by a 0?

jeremybz:
Please ignore this.  It is wrong.

Issue #1078, opened 2015-10-21
Sensu server does not send the extension name for checks 


franceesco:
Sensu server does not include the extension name in the payload when publishing checks to clients, the problem is in the embedded/lib/ruby/gems/2.0.0/gems/sensu-0.20.6/lib/sensu/server/process.rb file:
payload[:command] = check[:command] if check.has_key?(:command)
payload[:source] = check[:source] if check.has_key?(:source)
Here the extension should be added after the line 500 which adds the source:
payload[:extension] = check[:extension] if check.has_key?(:extension)
(Extension checks work only if the extension name and the check name is the same because the client uses check name as fallback when it doesn't find the extension field.)

franceesco:
The problem is the same with the extension options, this object isn't passed to the client either.
E.g.: the haproxy check extension (https://github.com/sensu/sensu-community-plugins/blob/master/extensions/checks/haproxy.rb) expects an object assigned to the "haproxy" key in the check object (e.g.: in line 186) which it never gets.
It would be nice if there was an optional "options" field in the check settings besides the name, command etc. attributes which would be sent to the client.

portertech:
@franceesco we will accept a PR to add extension to the request payload if it exists :+1:

gsalisbury:
I'd like to have a go at this if you're not already working on it @franceesco ?

franceesco:
@gsalisbury Feel free, I haven't had the time.

gsalisbury:
I've opened PR #1082 for this.

calebhailey:
Closing this as it looks like this was merged a long time ago! \o/ 

#monitoringlove

PR #1082, opened 2015-11-01
Send the extension name for checks

gsalisbury:
Fixes issue [#1078](https://github.com/sensu/sensu/issues/1078)

Sends the extension name if it exists in a check definition. Will output a warning message if both check :command and :extension keys exist. Unsure what behaviour would be best for that scenario, a command and extension shouldn't ever both be set so perhaps this would be better if it logged an error and didn't publish such checks.

From what I can see looking through the lib/sensu/client/process.rb process_check_request method, the extension name should be different from the check name, as the default behaviour is running the check as an extension check with the extension name set to the check name.

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -483,12 +483,12 @@ module Sensu
       end
 
       # Publish a check request to the transport. A check request is
-      # composted of a check `:name`, an `:issued` timestamp, and a
-      # check `:command` if available. The check request is published
-      # to a transport pipe, for each of the check `:subscribers` in
-      # its definition, eg. "webserver". JSON serialization is used
-      # when publishing the check request payload to the transport
-      # pipes. Transport errors are logged.
+      # composed of a check `:name`, an `:issued` timestamp, a check
+      # `:command` if available, and a check `:extension if available.
+      # The check request is published to a transport pipe, for each
+      # of the check `:subscribers` in its definition, eg. "webserver".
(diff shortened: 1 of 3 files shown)


lib/sensu/server/process.rb
```diff
@@ -498,6 +498,12 @@ def publish_check_request(check)
         }
         payload[:command] = check[:command] if check.has_key?(:command)
         payload[:source] = check[:source] if check.has_key?(:source)
+        if check.has_key?(:extension)
+          if check.has_key?(:command) 
+            @logger.warn("check has both command and extension options, command overrides extension")
```
portertech: Perhaps add some more context to the log event? e.g. `, :check => check`

portertech:
This PR is looking good :+1: Just that one minor adjustment I commented on and I believe we are good to go.

gsalisbury:
Ok thanks, have added that extra check context to the log event now.

Cheers,
George

portertech:
@gsalisbury thank you :+1: 

efd2987e-3867-11e4-9b56-e9145249548a

portertech:
Looks like we do not need the log warning https://github.com/sensu/sensu-settings/blob/master/lib/sensu/settings/validators/check.rb#L25-L26

portertech:
I am going to submit a PR to remove it, minor cleanup :+1:

portertech:
https://github.com/sensu/sensu/pull/1101

gsalisbury:
[image: :+1:]

Issue #1075, opened 2015-10-19
feature/add jit_origin to jit client data


roobert:
Currently there is no way to tell the origin of a check with `:source` defined. Adding a `jit_origin` parameter to check event data would add context to the check event data.

portertech:
:+1: Yes to this.

![1338](https://cloud.githubusercontent.com/assets/149630/10592241/ac75f37c-766f-11e5-8e61-0d8012a633e4.gif)

calebhailey:
Looks like this one fell of the radar, but it should be really easy to implement in 0.24.

roobert:
@calebhailey good stuff!

calebhailey:
I propose we call the attribute `origin`.

portertech:
#1267 was merged, adding this feature :+1:

PR #1093, opened 2015-11-09
set jit_origin to original client name when source parameter is set

roobert:
Related issue: https://github.com/sensu/sensu/issues/1075


portertech:
@roobert this will not work, as a JIT client is created only once, with no updates unless API POST /clients is used. jit_source needs to be in the check context, which also makes it easier to implement.

PR #1098, opened 2015-11-11
fix mutable filters bug

piavlo:
It was a :scream: :weary: bug, made filters much more than useless.

### DIFF
--- a/lib/sensu/server/filter.rb
+++ b/lib/sensu/server/filter.rb
@@ -258,7 +258,7 @@ module Sensu
       # @param callback [Proc]
       def event_filtered?(handler, event, &callback)
         if handler.has_key?(:filters) || handler.has_key?(:filter)
-          filter_list = handler.has_key(:filters)? handler[:filters].clone : [handler[:filter]]
+          filter_list = handler.has_key?(:filters)? handler[:filters].clone : [handler[:filter]]
           filter = Proc.new do |filter_list|
             filter_name = filter_list.shift
             if filter_name.nil?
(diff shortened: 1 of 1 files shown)


portertech:
@piavlo what is this?

```
[1] pry(main)> Array(nil || "foo")                                                                                                    
=> ["foo"]
[2] pry(main)> Array.new(nil || "foo")                                                                                                
TypeError: no implicit conversion of String into Integer
from (pry):2:in `initialize'
```

piavlo:
@portertech hmm i have only "filters" and no "filter" in my configs so i did not break on this
any way the issue with <code>Array(handler[:filters] || handler[:filter])</code>
is that this few lines later <code>filter_list.shift</code> mutates(empties) the 
@extensions[:handlers]['some_handler_with_filters'][:filters]
which effectively removes the filters from the handler and causes lots of pain

How about
<code>filter_list = handler.has_key(:filters)? handler[:filters].clone : [handler[:filter]]</code>
?

Btw this is also not that a user expects
<code>
[1] pry(main)> Array([] || "foo")     
=> []
[2] pry(main)> 
</code>
since "filters":[] is a valid config
https://github.com/sensu/sensu-settings/blob/master/lib/sensu/settings/validators/handler.rb#L89
and if user sets <b>"filter":[]</b> and <b>filter:"foo"</b> it will choose empty array and not merge these configs.

portertech:
@piavlo the plural has always taken precedence, e.g. handler vs handlers. We can make this more apparent in the documentation.

portertech:
@piavlo wouldn't a dup() be more appropriate?

portertech:
@piavlo I will merge this, as the tests do pass and `clone()` does work. I will likely make a minor change. 

Thank you for catching and fixing this one :+1: 

3ce7640e-44fd-11e4-9af4-2e3513179705

PR #1101, opened 2015-11-12
PR #1082 cleanup, config validation does not allow both check command and extension

portertech:
Removed command + extension warning log event as this situation cannot occur due to configuration validation (https://github.com/sensu/sensu-settings/blob/master/lib/sensu/settings/validators/check.rb#L25-L26). Minor test updates.


amdprophet:
:+1:

PR #1102, opened 2015-11-12
add support for api extensions

somic:
I would like to be able to add my own API endpoints/calls to sensu-api. Here I am proposing a simple mechanism to make it happen.

In a nutshell, make settings[:api_extensions] a hash with human readable name of extension as key and absolute file name as value.

```
api_extensions:
  my_extension1: /path/to/filename1.rb
  my_extension2: /path/to/filename2.rb
```

In filename*.rb it's as simple as:

```
Sensu::API::Process.class_eval do

  aget ...

end
```

Thoughts?

### DIFF
--- a/lib/sensu/api/process.rb
+++ b/lib/sensu/api/process.rb
@@ -49,6 +49,7 @@ module Sensu
             "Credentials" => "true",
             "Headers" => "Origin, X-Requested-With, Content-Type, Accept, Authorization"
           }
+          load_api_extensions
           on_reactor_run
           self
         end
@@ -92,6 +93,15 @@ module Sensu
         def test(options={})
           bootstrap(options)
           start
+        end
+
+        def load_api_extensions
+          if @settings[:api_extensions]
+            @settings[:api_extensions].each_pair do |name, filename|
+              settings.logger.info("loading api extension #{name} from #{filename}")
+              require filename
+            end
+          end
         end
       end
 


portertech:
@somic why not figure out how to support this via https://github.com/sensu/sensu-extension?

somic:
Interesting. Thanks! Let me take a look at that.

somic:
@portertech Just to confirm I understand. You are essentially talking about adding a new category to sensu extension (https://github.com/sensu/sensu-extension/blob/master/lib/sensu/extension/constants.rb) that can be included in sensu-api and provide a custom HTTP endpoint (like aget "/foo"), etc. yes?

portertech:
@somic yes, although we may need to add to the current extension API (Base) or add another class to inherit entirely.

kaos:
+1 for this in one shape or another.. :)

portertech:
@somic I am curious to hear where you got with this :-) As of Sensu 0.25, the API now uses EM HTTP Server, making it much easier to inject custom routes!

For example:

https://github.com/sensu/sensu/blob/master/lib/sensu/api/routes/clients.rb
https://github.com/sensu/sensu/blob/master/lib/sensu/api/routes.rb

somic:
Nice, thanks. I will take a look.

portertech:
Closing this, as the API has gone through some drastic changes since this PR was opened. We are not planning to implement this functionality, although it is something we would be happy to merge if done to work with the current API. Thanks :+1:

PR #1110, opened 2015-11-16
[WIP] sensu-install, only install sensu plugin gems if missing (version)

portertech:


amdprophet:
:+1:

NeckBeardPrince:
Awesome! Could be a dumb question but is this the method we should use to keep the sensu-plugin gem updated as well?

portertech:
@NeckBeardPrince installing the sensu plugins  with `sensu-install` will install the required version(s) of sensu-plugin.

PR #1119, opened 2015-11-26
Add client validation support

### LINKED ISSUE #1118: Client Verification/Security
Hi there,

I'm wondering if there's a way to to verify that the data sent by a client (results, keepalives) are coming from the right client. I just performed a quick test, and it appears that if a client system is compromised, an attacker can gain access to the RabbitMQ connection, and inject messages onto the queue disguising itself as another client.

For example, I have here `test-app01`, and using [pika](https://pika.readthedocs.org/en/0.10.0/), I'm able to enqueue a message for `test-app02`:

```python
#!/usr/bin/env python
import pika

connection = pika.BlockingConnection(pika.URLParameters('amqps://sensu:sensu@server-ip/%2Fsensu?ssl_options=%7B%27keyfile%27%3A%20%27%2Fetc%2Fsensu%2Fssl%2Fsensu_client_key.pem%27%2C%20%27certfile%27%3A%20%27%2Fetc%2Fsensu%2Fssl%2Fsensu_client_cert.pem%27%7D'))
channel = connection.channel()

channel.basic_publish(exchange='',
                      routing_key='results',
                      body='{"client":"test-app02","check":{"name":"test","issued":1448422108,"command":"/etc/sensu/plugins/check-procs.rb -p test","executed":1448422108,"duration":0.004,"output":"I am a hacker","status":0}}')
```

In the example above, I was able to successfully inject this message, and the `sensu-server` happily processed it without question:

```json
{"timestamp":"2015-11-25T11:23:01.933266+0000","level":"debug","message":"received result","result":{"client":"test-app02","check":{"name":"test","issued":1448422108,"command":"/etc/sensu/plugins/check-procs.rb -p test","executed":1448422108,"duration":0.004,"output":"I am a hacker","status":0}}}
{"timestamp":"2015-11-25T11:23:01.933573+0000","level":"debug","message":"processing result","result":{"client":"test-app02","check":{"name":"test","issued":1448422108,"command":"/etc/sensu/plugins/check-procs.rb -p test","executed":1448422108,"duration":0.004,"output":"I am a hacker","status":0}}}
{"timestamp":"2015-11-25T11:23:01.934856+0000","level":"debug","message":"storing check result","check":{"command":"/etc/sensu/plugins/check-procs.rb -p test","subscribers":["production","test"],"interval":60,"name":"test","issued":1448422108,"executed":1448422108,"duration":0.004,"output":"I am a hacker","status":0}}
```

Using RabbitMQ ACLs and SSL certificates won't necessarily help in this case if the server compromised was already a trusted client. The `result` and `keepalive` queues must be writable by all clients according to https://github.com/sensu/sensu/issues/696. By gaining access to the local client configuration, I am able to disguise myself as any other trusted client and taint their results.

hany:
- Fixes #1118
- Needs https://github.com/sensu/sensu-transport/pull/21
- Utilizes RabbitMQ's Validated User-ID feature to implement client
  validation.
- An administrator can configure unique RabbitMQ credentials per client,
  with the username set to the hostname of the client. This should match
  the client "name" configuration option of the sensu-client.
- The sensu-client (and sensu-transport) can be configured to add a
  user_id attribute to each published message. This is probably good
  practice to have in any case.
-  Check for the existence of the user_id header in each message, and
  reject those that don't have it set.
-  Check to ensure that the "client" field of the message payload matches
  the user_id header set.

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -449,7 +449,26 @@ module Sensu
           begin
             result = MultiJson.load(message)
             @logger.debug("received result", :result => result)
-            process_check_result(result)
+            if message_info.attributes.has_key?(:user_id)
+              if message_info.attributes[:user_id] == @settings[@settings[:transport][:name]][:user] &&
+                   result.has_key?(:check) &&
+                   result[:check][:name] == "keepalive"
+
+                @logger.debug("keepalive from server")
+                process_check_result(result)
+              elsif message_info.attributes[:user_id] == result[:client]
+                @logger.debug("client validation passed", result[:client])
+                process_check_result(result)
+              else
+                @logger.error("client validation error - user_id mismatches client name",
+                  :payload => message
+                )
+              end
+            else
+              @logger.error("client validation error - user_id missing from message",
+                :payload => message
+              )
+            end
           rescue MultiJson::ParseError => error
             @logger.error("failed to parse result payload", {
               :message => message,


hany:
I have not implemented a server-side option to enable this feature. Not sure where such an option would go. Any thoughts on this?

hany:
This is no longer needed and has been replaced with #1119.

PR #1120, opened 2015-11-29
Make the client socket accept non-ASCII characters

miyakawataku:
JSON data encoded as UTF-8 may contain non-ASCII bytes (>=0x80), but currently the client socket rejects such data. This pull-request makes the client socket accept non-ASCII characters and parse them as UTF-8.

Reference:

- [Character Encoding in JSON](https://tools.ietf.org/html/rfc7159#section-8.1)

Indeed non-ASCII characters can be encoded using `\uXXXX` notation, but it imposes extra effort on the client side. Particularly, ruby code can be messy because popular JSON libraries do not support `\uXXXX` notation in dump methods. Example:

- [out_sensu.rb in fluent-plugin-sensu](https://github.com/miyakawataku/fluent-plugin-sensu/blob/v1.0.1/lib/fluent/plugin/out_sensu.rb#L200)

### DIFF
--- a/lib/sensu/client/socket.rb
+++ b/lib/sensu/client/socket.rb
@@ -186,6 +186,7 @@ module Sensu
       #
       # @param [String] data to be processed.
       def process_data(data)
+        data = data.force_encoding('utf-8')
         if data.strip == "ping"
           @logger.debug("socket received ping")
           respond("pong")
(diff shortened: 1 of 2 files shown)


portertech:
@miyakawataku Thank you :+1: 

Can we add a test to produce an encoding error when parsing to ensure everything hums along?

miyakawataku:
I think it would be preferable to log a warning for an encoding error, and then accept the check. I think the check should not be dropped even if it contains an encoding error, because we want to receive notification for the check anyway. What do you think about it @portertech?

And yes, I will add a warning log for encoding errors if it's OK.

miyakawataku:
Added warn-logging for encoding error. Please check it out @portertech.

portertech:
@miyakawataku thank you. I am going to merge this, it should make it into the next release (0.22) unless I discover an valid reason why we forced ASCII in the past.

PR #1125, opened 2015-12-03
Add simple client validation

hany:
Clients can now optionally set a `client_key` in their configuration,
which will be passed along in keepalives and check results. If set,
validation is as follows:

- When new clients are registered, the `client_key` is stored along with
  all the other attributes in Redis.
- When keepalives or check results are received, the respective
  functions will call `validate_client` to ensure the messages are
  authentic.
- `validate_client` will check the registry for the client and fetch the
  `client_key`.
  - If a previously registered client had a `client_key`, and the
    incoming message did not, it will fail validation.
  - If the keys do not match, it will fail validation.
  - If the registered client did not have a `client_key`, it will pass
    validation.
  - If there was no previously registered client, it will pass
    validation.

This is a simplistic form of validation and should not be considered
100% secure. This merely adds some basic security to incoming messages.

### DIFF
--- a/lib/sensu/client/process.rb
+++ b/lib/sensu/client/process.rb
@@ -74,10 +74,11 @@ module Sensu
       end
 
       # Publish a check result to the transport for processing. A
-      # check result is composed of a client (name) and a check
-      # definition, containing check `:output` and `:status`. JSON
-      # serialization is used when publishing the check result payload
-      # to the transport pipe. Transport errors are logged.
+      # check result is composed of a client (name), the unique
+      # client_key, and a check definition, containing check
+      # `:output` and `:status`. JSON serialization is used when
+      # publishing the check result payload to the transport pipe.
+      # Transport errors are logged.
       #
       # @param check [Hash]
       def publish_check_result(check)
@@ -85,6 +86,9 @@ module Sensu
           :client => @settings[:client][:name],
           :check => check
         }
+        if @settings[:client].has_key?(:client_key)
+          payload.merge(:client_key => @settings[:client][:client_key])
+        end
         @logger.info("publishing check result", :payload => payload)
         @transport.publish(:direct, "results", MultiJson.dump(payload)) do |info|
           if info[:error]
--- a/lib/sensu/client/process.rb
+++ b/lib/sensu/client/process.rb
@@ -84,11 +84,9 @@ module Sensu
       def publish_check_result(check)
         payload = {
           :client => @settings[:client][:name],
+          :client_key => @settings[:client][:client_key],
           :check => check
         }
-        if @settings[:client].has_key?(:client_key)
-          payload.merge!(:client_key => @settings[:client][:client_key])
-        end
         @logger.info("publishing check result", :payload => payload)
         @transport.publish(:direct, "results", MultiJson.dump(payload)) do |info|
           if info[:error]
--- a/lib/sensu/client/process.rb
+++ b/lib/sensu/client/process.rb
@@ -87,7 +87,7 @@ module Sensu
           :check => check
         }
         if @settings[:client].has_key?(:client_key)
-          payload.merge(:client_key => @settings[:client][:client_key])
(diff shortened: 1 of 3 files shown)


hany:
Fixes #1118

runningman84:
any news here?

hany:
Hey Folks, seeing how this is a security issue, it would be great to have a set of eyes as soon as possible on this. As per #1118, it was demonstrated that a single compromised host can masquerade as any other Sensu-monitored host by simply guessing the client name. In addition, if a particular Sensu installation makes use of handlers that take action, this security issue can potentially be used to cause harm (think handlers that restart services/servers).

This PR takes a middle-of-the-road approach, and is actually in line with the discussion in #1118. Feedback and quick motion would be appreciated, and in this user's opinion, prudent.

lib/sensu/client/process.rb
```diff
@@ -85,6 +86,9 @@ def publish_check_result(check)
           :client => @settings[:client][:name],
```
portertech: Should just add `:client_key => @settings[:client][:client_key]`, which will just be `nil` by default. We can then check for `nil?()` when processing.

lib/sensu/server/process.rb
```diff
@@ -55,20 +90,29 @@ def update_client_registry(client, &callback)
...
+      # Transport message acknowledgements are used to ensure the client
+      # registry is updated successfully. Keepalive JSON parsing errors are
+      # logged.
       def setup_keepalives
         @logger.debug("subscribing to keepalives")
         @transport.subscribe(:direct, "keepalives", "keepalives", :ack => true) do |message_info, message|
           @logger.debug("received keepalive", :message => message)
           begin
             client = MultiJson.load(message)
-            update_client_registry(client) do
```
portertech: Should move `validate_client()` within `update_client_registry()` to keep `setup_keepalives()` simple and easy to follow.

portertech: We should probably utilize `retrieve_client()` in `validate_client()`.

lib/sensu/server/process.rb
```diff
@@ -578,11 +630,14 @@ def setup_check_request_publisher
       #
       # @param client_name [String]
       # @param check [Hash]
-      def publish_check_result(client_name, check)
+      def publish_check_result(client_name, client_key, check)
         payload = {
           :client => client_name,
```
portertech: We can add `:client_key => client_key`, which will default to nil.

portertech:
I made a few comments. This is a great base :+1:

portertech:
`client_key` is an opt-in feature, which will naturally have a performance penalty as results with a `client_key` that is not `nil` will require validate.

hany:
@portertech thanks a lot for your feedback! I've made a commit that addresses your comments. Please note that I did not utilize `retrieve_client()` in `validate_client()` as that function typically expects a specifically formatted "result":

```ruby
client_key = result[:check][:source] || result[:client]
```

However, `validate_client()` is written to accept both keepalives and results from clients *and* the server:

```ruby
client_name = client[:name] || client[:client]
```

We'd likely need to refactor `retrieve_client()` to be more generic and accept multiple different forms, but it seems a bit unnecessary and potentially complex.

As for the performance impact, the cost is a single extra Redis call. Doesn't seem to severe for the security gained.

Lastly, the broken Travis build does not appear related to this commit:

https://travis-ci.org/sensu/sensu/jobs/96882430#L138

hany:
Hey @portertech any update on this issue? Would love to see this move forward, especially considering the sensitive security-related nature of this.

hany:
Hey folks, @portertech, any update on this front?

portertech:
@hany I created https://github.com/sensu/sensu/pull/1154, an alternative method of achieving the same effect, with a few performance improvements.

portertech:
Closing this as #1154 has been reviewed and merged. Thank you for doing the hard work :+1:

PR #1141, opened 2015-12-23
Client registration events

portertech:
This PR allows for client registration event creation (and handling) when a sensu-client is first added to the registry. By default, client registration events use a handler named "registration". Clients can configure their own registration handler, e.g. `"registration": {"handler": "foo"}`.

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -37,6 +37,15 @@ module Sensu
         @handling_event_count = 0
       end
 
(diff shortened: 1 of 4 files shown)


amdprophet:
:+1:

agoddard:
:+1: I like this. Custom handlers are then available to do anything further (silence client on registration, run once-off check at registration, etc) :shipit:

portertech:
shaq-and-cat-gif

calebhailey:
\o/

PR #1154, opened 2016-01-19
Support client signatures

portertech:
This feature & PR is based on the work done by @hany (https://github.com/sensu/sensu/pull/1125). 

Sensu clients can optionally set a `"signature"` in their definitions. A client signature is a unique string identifier, which is used to sign transport messages, for message source verification. Once a client signature is configured/set, the registered signature cannot be modified. A client must be deleted from the registry (via the API) in order to modify an existing client signature.

If a client has a signature, all keepalives and check results that it publishes must be signed with the correct signature or they will be discarded. This is not a replacement for other security measures.


portertech:
Need sensu-settings validation.

amdprophet:
:+1:  Looks good & transport agnostic.

PR #1171, opened 2016-02-22
Performance improvements

portertech:
This PR reduces method block argument usage in most core libraries, using `yield()` instead of `callback.call()`. This PR also includes RabbitMQ transport error handling improvements.


agoddard:
getinsertpic.com

agoddard:
:+1:

Issue #1177, opened 2016-03-01
Update to 0.22.0 - sensu server crashes with "update_client_registry': undefined method `empty?' for nil:NilClass (NoMethodError)" 


alex-leonhardt:
Hi,

since updating to 0.22.x we've noticed several sensu server crashes with always the same message - this seems directly related to the update as we've never had this happening on 0.19.x 

```
{"timestamp":"2016-03-01T06:22:13.887513+0000","level":"warn","message":"reconnecting to redis"}
/opt/sensu/embedded/lib/ruby/gems/2.2.0/gems/sensu-0.22.0/lib/sensu/server/process.rb:119:in `block (2 levels) in update_client_registry': undefined method `empty?' for nil:NilClass (NoMethodError)
	from /opt/sensu/embedded/lib/ruby/gems/2.2.0/gems/em-redis-unified-1.0.1/lib/em-redis/redis_protocol.rb:487:in `call'
	from /opt/sensu/embedded/lib/ruby/gems/2.2.0/gems/em-redis-unified-1.0.1/lib/em-redis/redis_protocol.rb:487:in `dispatch_response'
	from /opt/sensu/embedded/lib/ruby/gems/2.2.0/gems/em-redis-unified-1.0.1/lib/em-redis/redis_protocol.rb:432:in `process_cmd'
	from /opt/sensu/embedded/lib/ruby/gems/2.2.0/gems/em-redis-unified-1.0.1/lib/em-redis/redis_protocol.rb:406:in `receive_data'
	from /opt/sensu/embedded/lib/ruby/gems/2.2.0/gems/eventmachine-1.0.8/lib/eventmachine.rb:193:in `run_machine'
	from /opt/sensu/embedded/lib/ruby/gems/2.2.0/gems/eventmachine-1.0.8/lib/eventmachine.rb:193:in `run'
	from /opt/sensu/embedded/lib/ruby/gems/2.2.0/gems/sensu-0.22.0/lib/sensu/server/process.rb:23:in `run'
	from /opt/sensu/embedded/lib/ruby/gems/2.2.0/gems/sensu-0.22.0/exe/sensu-server:10:in `<top (required)>'
	from /opt/sensu/bin/sensu-server:23:in `load'
	from /opt/sensu/bin/sensu-server:23:in `<main>'
```

It'd be great to find out what is causing this.

Thanks!
Alex

portertech:
[...]

PR #1178, opened 2016-03-02
client signature may return nil

portertech:
This PR fixes https://github.com/sensu/sensu/issues/1177


agoddard:
getinsertpic.com

Issue #1182, opened 2016-03-04
A TTL value submitted as a string can cause the sensu-server to die.


jbehrends:
We recently had a sensu-server die for a few days unknowingly. The last log entry before dieing was this:

```
{"timestamp":"2016-03-04T09:45:18.687212-0800","level":"warn","message":"reconnecting to redis"}
/opt/sensu/embedded/lib/ruby/gems/2.0.0/gems/sensu-0.20.1/lib/sensu/server/process.rb:672:in `>=': comparison of Fixnum with String failed (ArgumentError)
	from /opt/sensu/embedded/lib/ruby/gems/2.0.0/gems/sensu-0.20.1/lib/sensu/server/process.rb:672:in `block (5 levels) in determine_stale_check_results'
	from /opt/sensu/embedded/lib/ruby/gems/2.0.0/gems/em-redis-unified-1.0.0/lib/em-redis/redis_protocol.rb:489:in `call'
	from /opt/sensu/embedded/lib/ruby/gems/2.0.0/gems/em-redis-unified-1.0.0/lib/em-redis/redis_protocol.rb:489:in `dispatch_response'
	from /opt/sensu/embedded/lib/ruby/gems/2.0.0/gems/em-redis-unified-1.0.0/lib/em-redis/redis_protocol.rb:436:in `process_cmd'
	from /opt/sensu/embedded/lib/ruby/gems/2.0.0/gems/em-redis-unified-1.0.0/lib/em-redis/redis_protocol.rb:408:in `receive_data'
	from /opt/sensu/embedded/lib/ruby/gems/2.0.0/gems/eventmachine-1.0.3/lib/eventmachine.rb:187:in `run_machine'
	from /opt/sensu/embedded/lib/ruby/gems/2.0.0/gems/eventmachine-1.0.3/lib/eventmachine.rb:187:in `run'
	from /opt/sensu/embedded/lib/ruby/gems/2.0.0/gems/sensu-0.20.1/lib/sensu/server/process.rb:23:in `run'
	from /opt/sensu/embedded/lib/ruby/gems/2.0.0/gems/sensu-0.20.1/bin/sensu-server:10:in `<top (required)>'
	from /opt/sensu/bin/sensu-server:23:in `load'
	from /opt/sensu/bin/sensu-server:23:in `<main>'
```

Looking at the code, the line it's complaining about is this:
`if time_since_last_execution >= check[:ttl]`

[...]

PR #1183, opened 2016-03-04
Handle checks with non integer ttl values

skurylo:
fixes sensu/sensu#1182

If a user accidentally injects a string ttl value instead of an integer
it will crash the sensu server process.


skurylo:
I am investigating why this works locally but fails on travis

portertech:
@skurylo alternatively, we could guard against an invalid TTL with input validation https://github.com/sensu/sensu/blob/master/lib/sensu/client/socket.rb#L121

portertech:
This Sensu client socket now guards against invalid check TTL values :+1:

PR #1191, opened 2016-03-15
deregister event

zbintliff:
Wanted to move the deregister event inside of the sensu client instead of the init script.  That way its easy to configure via the puppet module.

Since the register event is also configured via the client, I think it makes more sense to have both live inside of the code.

### DIFF
--- a/lib/sensu/client/process.rb
+++ b/lib/sensu/client/process.rb
@@ -381,6 +381,23 @@ module Sensu
         end
       end
 
+      # Sends a deregister event to be processed by sensu server
+      def send_deregister_event
+        if @settings[:client][:deregister_handler].nil?
+          @logger.warn("deregister enabled but no handler passed, not sending event")
+        else
+          @logger.info("sending deregister event to #{@settings[:client][:deregister_handler]}")
+          check = {
+            :name => 'deregister',
+            :output => 'delete client as a result of clean shutdown',
+            :process => 'init',
+            :status => 1,
+            :handler => @settings[:client][:deregister_handler]
+          }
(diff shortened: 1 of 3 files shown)


zbintliff:
Hopefully this way is preferred to sending the event via the init script: https://github.com/sensu/sensu-build/blob/master/sensu_configs/init.d/sensu-service#L200-L204

zbintliff:
Currently set up as `{"client":{"deregister":true,"deregister_handler":"HANDLER}}`  but can make it match registration with `{"client":"deregister":{"handler":"HANDLER"}}`

lib/sensu/client/process.rb
```diff
@@ -455,9 +472,11 @@ def resume
       # Stop the Sensu client process, pausing it, completing check
       # executions in progress, closing the transport connection, and
       # exiting the process (exit 0). After pausing the process, the
-      # process/daemon `@state` is set to `:stopping`.
+      # process/daemon `@state` is set to `:stopping`.  Also sends
+      # deregister event if configured to do so.
       def stop
         @logger.warn("stopping")
+        send_deregister_event if @settings[:client][:deregister] == true
         pause
```
zbintliff: Wasn't sure about the timing here.  Might be best to put the deregister event after pause so we have a 'guarantee' that no more results/events will be sent.

cwjohnston:
Hi @zbintliff, thanks for the PR! I've proposed #1309 to supersede this PR, adding support for client `deregistration` attributes to be merged into the resulting event.

zbintliff:
Ok great. I'll close this out then.

PR #1198, opened 2016-03-24
Support token substitution in filter eval attributes

portertech:
Possible solution for: https://trello.com/c/QDqdU5kk/79-new-support-for-filter-eval-token-substitution

A top level `"eval":` setting tends to produce Ruby eval() strings that are long and difficult to read, as conditionals grow, e.g. `&& ||`. This PR adds token support to the existing filter eval attribute feature, promoting more concise filter definitions.

Example:

```
"attributes": {
  "occurrences": "eval: value == :::check.occurrences::: || value % 60 == 0"
}
```

### DIFF
--- a/lib/sensu/server/filter.rb
+++ b/lib/sensu/server/filter.rb
@@ -167,6 +167,16 @@ module Sensu
         end
       end
 
+      # Process a filter eval attribute, a Ruby `eval()` string
+      # containing an expression to be evaluated within the
+      # scope/context of a sandbox. This methods strips away the
+      # expression prefix, `eval:`, and substitues any dot notation
+      # tokens with the corresponding event data values. If there are
+      # unmatched tokens, this method will return `nil`.
+      #
+      # @event [Hash]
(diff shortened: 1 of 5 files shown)


lib/sensu/server/filter.rb
```diff
@@ -165,28 +167,63 @@ def handle_severity?(handler, event)
...
-          @logger.error("filter attribute eval error", {
-            :raw_eval_string => raw_eval_string,
-            :value => value,
-            :error => error.to_s
-          })
+      def eval_attribute_value(event, raw_eval_string, raw_value)
+        eval_string = process_eval_string(event, raw_eval_string)
+        unless eval_string.nil?
+          begin
+            value = Marshal.load(Marshal.dump(raw_value))
```
portertech: Marshal load()/dump() is used to deep dup any objects that are to be provided via `value`.

calebhailey:
Assuming the tests pass, this LGTM!

calebhailey:
cantwait

PR #1200, opened 2016-03-29
Set the check interval to 30s for stale ttl events

solarkennedy:
At Yelp we make heavy use of the TTL stale-check feature.

We also use advanced filtering logic based on the config of the check.

The problem comes when there is a mismatch between the original configuration of a check (maybe it runs every 24hs), and when it run in "stale ttl mode" (which is 30s)

This patch makes it so the stale ttl thing correctly sets the "real" interval when it is in this mode, which will give our handlers better hints about how frequently to anticipate this event.

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -953,6 +953,7 @@ module Sensu
                   client_name = result_key.split(":").first
                   keepalive_event_exists?(client_name) do |event_exists|
                     unless event_exists
+                      check[:interval] = 30
                       check[:output] = "Last check execution was "
                       check[:output] << "#{time_since_last_execution} seconds ago"
                       check[:status] = 1


portertech:
@solarkennedy I am concerned that this change will create confusion, as it would then be impossible to determine the actual check interval.

solarkennedy:
Well, when a check goes stale the *actual* check interval *is* 30s.

When whatever created the check perks up next time, it should reset the interval to the *original* interval. (right?)

solarkennedy:
I've rebased. @portertech In a different sense, some of the yelp code *is* already confused because the check interval says one thing (24hs say) but in ttl mode at this point in the code, the *actual* interval is 30s. I would prefer that the result data reflect what the actual check interval is IRL, not what it was originally set to.

CVTJNII:
Oops, I didn't see this and duplicated it in #1514, which I'm leaving open as I added tests there.

I'm with @solarkennedy on this, having the send interval not match the reported interval is a bad thing, worse than the interval changing as interval is used to calculate re-alert counts.  I think these are symptoms of a larger and more fundamental issue in TTL handling, if we want the interval to not change then the TTL needs to fire at the same interval as the check, which is complicated.  Based on @portertech's comment in #1302 and simply how much harder that will be to manage I don't think that's going to be fixed in the near term.  So I think this is a good stopgap solution to mitigate real-world problems where TTL alerts are sent when they shouldn't.  As for confusion, well, sadly it's confusing no matter what if the TTL interval doesn't match the check.

> When whatever created the check perks up next time, it should reset the interval to the original interval. (right?)
Yes, it will.

solarkennedy:
@CVTJNII gotcah. I'll close this as dupe as yours is a bit more comprehensive.

Issue #1025, opened 2015-08-14
Sensu does not set event[:check][:type] = "standard" for standard checks.


erenfro:
Checks with type "metric" specifically set, always has that returned in the json/hash of the event object, but standard checks, which are the default, unless specifically defined in the client config, do not specifically provide the value for event[:check][:type], and it should be "standard".

This causes issues when trying to handle standard and metric checks, and any other kind of future checks that may eventually be implemented. With the way ruby hash's work, the only viable method I found to determine specifically if a check is standard is with the following code:

          if event[:check].key?(:type)
            check_type = event[:check][:type]
          else
            check_type = "standard"
          end

Which.. Honestly, isn't very good to assume everything is standard.

portertech:
:+1: we should set the default to either "standard" or "service" for defined context.

portertech:
This is an excellent question for the next Sensu community survey.

calebhailey:
Tagging this for 0.24.

calebhailey:
Related: https://github.com/sensu/sensu/issues/990

PR #1266, opened 2016-05-17
Result check type defaults to "standard"

portertech:
Closes https://github.com/sensu/sensu/issues/1025


calebhailey:
Nice one-liner 👍

PR #1267, opened 2016-05-17
Set event check origin to client name if check source set

portertech:
Close https://github.com/sensu/sensu/issues/1075


calebhailey:
One-liners are the best 👍

Issue #1281, opened 2016-05-25
Token-substitution should apply to all attributes of a check


UnitedMarsupials:
Currently token-substitution is only applied to the `command`-attribute. This makes sense, because `command` is the only place, where arbitrary strings can be encountered.

The only place among _standard_ attributes... However, there can be non-standard ones attached to a check. For example, I'm trying to add a `Chart`-attribute, that would include a link to Graphite-produced image:

```json
		"NTP": {
			"command":	"check-ntp.rb  -w 2 -c 5 -s 4",
			"interval":	3677,
			"Chart":	"http://graphite.example.net/render/?width=512&height=360&from=-24hours&lineWidth=4&target=stats.:::name:::.ntpstats.offset&vtitle=NTP%20offset%20(milliseconds)&fname/NTP-offset-:::name:::.png",
			"subscribers":	[ "unix" ]
		}
```

It almost works -- the image is nicely embedded in Uchiwa's check-details:

![Uchiwa chart sample](http://aldan.algebra.com/~mi/tmp/uchiwa-chart-example.png)

Unfortunately, the image contains no data, because the `:::name:::` tokens weren't processed by Sensu. I can, probably, write my own mutator to do it, but the right place for it seems to be in the `lib/sensu/client/process.rb`: loop through all of the `check`'s fields and perform substitution on them before `publish_check_results`. 

Possibly skipping the standard fields (`interval`, `subscribers`, `handler`). The actual coding seems simple enough -- would it be accepted, if I did it?

UnitedMarsupials:
Actually, perhaps, the right place for the token-substitution is not so obvious... Currently the client performs it _every time the check runs_. Would not it make more sense to perform the substitution _once_ per check -- whenever one is either received from the server or loaded locally?

Are there tokens today, that change value from one check-execution to another?

portertech:
@UnitedMarsupials I think this idea is great :+1: applying token substitution to any string check attribute value (top-level only) is appropriate, e.g. `next unless value.is_a?(String)`. Your proposed changes look pretty good, however, the Sensu code base avoids the use of `return` when any async code is present, hence the conditional wrapping. We should probably use string interpolation instead of concatenation, e.g. `"Unmatched '#{key}' tokens: #{unmatched_tokens.join(", ")}"`.

portertech:
[...]

PR #1290, opened 2016-05-26
Perform token-substitution in all String-attributes of a check

UnitedMarsupials:
See #1281 . Thanks!

### DIFF
--- a/lib/sensu/client/process.rb
+++ b/lib/sensu/client/process.rb
@@ -99,7 +99,7 @@ module Sensu
         end
       end
 
-      # Perform token-substitution for a object. Strings
+      # Perform token-substitution for an object. Strings
       # are passed to substitute_tokens(), arrays and sub-hashes are
       # processed recursively. Numbers (and other types?) are skipped
       # at the moment, because substitute_tokens() can not handle them
@@ -122,11 +122,11 @@ module Sensu
             value, unmatched = obj_substitute_tokens(value)
             unmatched_tokens.push(*unmatched)
             value
-	  }
+          }
         when String
           obj, unmatched_tokens = substitute_tokens(obj, @settings[:client])
         end
-	unmatched_tokens.uniq! if unmatched_tokens
+        unmatched_tokens.uniq! if unmatched_tokens
         [obj, unmatched_tokens]
       end
 
--- a/lib/sensu/client/process.rb
+++ b/lib/sensu/client/process.rb
@@ -99,6 +99,37 @@ module Sensu
         end
       end
 
+      # Perform token-substitution for a object. Strings
+      # are passed to substitute_tokens(), arrays and sub-hashes are
+      # processed recursively. Numbers (and other types?) are skipped
+      # at the moment, because substitute_tokens() can not handle them
+      # anyway.
+      #
+      # @param	[Object]
+      # @return	[Object] updated object of the same type as the argument
+      # 	plus accumulated list of unmatched tokens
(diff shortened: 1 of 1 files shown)


UnitedMarsupials:
Sorry, seems like my editor inserted a tab in one line instead of 8 spaces as seems to be Sensu's coding style. The comment should, probably, also be edited to say "a**n** object".

But I don't want to commit these decorative changes myself to avoid jeopardizing this feature making it into 0.24 as promised by @portertech in #1281 :-) Thanks!

lib/sensu/client/process.rb
```diff
@@ -99,6 +99,37 @@ def publish_check_result(check)
...
+      # Perform token-substitution for a object. Strings
+      # are passed to substitute_tokens(), arrays and sub-hashes are
+      # processed recursively. Numbers (and other types?) are skipped
+      # at the moment, because substitute_tokens() can not handle them
+      # anyway.
+      #
+      # @param	[Object]
+      # @return	[Object] updated object of the same type as the argument
+      # 	plus accumulated list of unmatched tokens
+      def obj_substitute_tokens(obj)
```
portertech: Let's drop `obj_` from the method name :+1:

lib/sensu/client/process.rb
```diff
@@ -99,6 +99,37 @@ def publish_check_result(check)
...
+          obj.map! {|value|
+            value, unmatched = obj_substitute_tokens(value)
+            unmatched_tokens.push(*unmatched)
+            value
+	  }
+        when String
+          obj, unmatched_tokens = substitute_tokens(obj, @settings[:client])
+        end
+	unmatched_tokens.uniq! if unmatched_tokens
+        [obj, unmatched_tokens]
```
portertech: Just do `unmatched_tokens.uniq` here?

lib/sensu/client/process.rb
portertech: Maybe `obj` -> `object` or `definition`?

portertech:
@UnitedMarsupials need the PR to target the release/0.24 branch.

lib/sensu/client/process.rb
```diff
@@ -99,6 +99,37 @@ def publish_check_result(check)
...
+        case obj
+        when Hash
+          unmatched_tokens = []
+          obj.each do |key, value|
+            obj[key], unmatched = obj_substitute_tokens(value)
+            unmatched_tokens.push(*unmatched)
+          end
+        when Array
+          unmatched_tokens = []
+          obj.map! {|value|
```
portertech: Keep to the `do` `end` style that  you used for Hash?

UnitedMarsupials:
> Let's drop obj_ from the method name

But then it will conflict with the original `substitute_tokens` from `utilities.rb`, no?

> Just do unmatched_tokens.uniq here?

Where?

> need the PR to target the release/0.24 branch.

Sorry, I don't understand... Oh, you mean, pull-request must be against 0.24, rather than master branch? Ok...

UnitedMarsupials:
How about now? The `obj_` prefix is still there -- open to suggestions...

UnitedMarsupials:
Maybe, this new method of mine should replace the old, string-only `substitute_tokens`? Incorporating the old code inside its `when String` branch. But then it will also apply to the filter-handling -- which may or may not be desirable...

portertech:
Merged https://github.com/sensu/sensu/pull/1294 :+1:

Issue #1282, opened 2016-05-25
Option to suppress expired TTL events when keepalive is failing


nhooey:
Currently when a check hasn't been run within the configured `TTL`, an event is generated with output: `Last check execution was X seconds ago`.

When a node is shut down, or otherwise disconnected from Sensu, a check result is generated for each check that hasn't been run within the `TTL`.

This information isn't really useful when the `keepalive` check failing is the real issue, so it would be helpful to be able to suppress expired `TTL` events when the `keepalive` check is failing.

mattyjones:
I agree this is frustrating. I have a handler for dealing with this but I am waiting on a few other parts. A native solution would be best.

nhooey:
@mattyjones Could you publish this handler so I can use it?

Before you do that, a couple of questions:

1. Is it a regular handler or an extension handler?
1. Does the handler make an API request to check the keepalive state of the node in question?

nhooey:
Is it possible to create a filter that runs code to check if a client is currently failing a keepalive check?
@portertech, @calebhailey

nhooey:
I went ahead and made an [TTL Keepalive Filter](https://github.com/nhooey/sensu-plugins-ttl-keepalive-filter/blob/master/ttl_keepalive_filter.rb) that is an extension filter.

It filters out failed results due to expired TTL, where the client is currently failing its keepalive check (the overarching issue). I just tested it in production and it works, with decent performance.

It also depends on my pull request: #1291

@mattyjones, @portertech, @calebhailey

calebhailey:
Tagging this for consideration with the 0.25 release.

selbyk:
+1

calebhailey:
The 0.25 release is going to hit today with fewer features than originally anticipated & primarily only some internal improvements. Moving this issue to 0.26.

RyanJarv:
This seems to be a problem for us when Sensu thinks everything is down, the TTL event's push us over what our sensu servers can currently handle adding to the delay when actual results are processed. Running a flushall on redis and clearing the queues get's things running again as a temporary solution.

calebhailey:
Related: https://github.com/sensu/sensu/issues/861#issuecomment-130136052

portertech:
I merged https://github.com/sensu/sensu/pull/1428 0.26 will have this functionality :+1:

PR #1291, opened 2016-05-27
Allow extension handlers to have `filters`.

nhooey:


nhooey:
#1282 depends on this change.

nhooey:
With this change, an extension handler can defined its own instance method to return `filter` or `filters`, but this issue needs to be solved to get it from configuration: sensu/sensu-settings#30.

portertech:
@nhooey Sensu handler extensions do support filters. Add "filter" or "filters" to `definition()`.

nhooey:
What do you mean by `definition()`? Could you be more specific?

Is `definition()` a member function of the Extension handler class? Is `definition()` a section in a `handler` definition in the JSON configuration?

Since my handler is an extension handler, it's not defined in the configuration JSON, only mentioned in the `handlers` section of each `check`.

nhooey:
For anyone else looking at this, it's the `definition()` member function of the extension handler where the `filters` entry is set. So my pull request isn't needed.

Issue #1215, opened 2016-04-08
Sensu API never starts listening when transport never successfully connects


nhooey:
When the Sensu API is started up, the debug message: `connecting to transport` prints, and if the connection fails for a more subtle reason than `Connection refused`, Sensu constantly attempts to reconnect every 2 seconds, without the `Transport.connect()` ever yielding a `connection` object.

On startup when the `transport` connection fails, this prevents the API from ever listening.

This problem occurs with Sensu 0.23 from the Ubuntu repository, but in Sensu 0.22, the API successfully listens and tells Uchiwa that the transport is not connected, which is the correct behaviour since my transport is currently not working for some configuration reason that's my fault.

The log messages for this are also pretty terrible, as Sensu API prints nothing at the default log level. It should print the following with `@logger.{info,error}()` by default:

- Connecting to transport
- Connection failed (including the reason), reconnecting to transport

Also, the API should still come up and serve requests even when the `transport` fails to connect, so clients like Uchiwa can see the reason why the API is in a bad state.

See my notes on these commits for a possible solution:

- [sensu/sensu/commit/7c39541a](https://github.com/sensu/sensu/commit/7c39541a#diff-2a7cdd2fd11ac77dc3f35be73e06b7d0R188)
- [sensu/sensu-transport/commit/ddafe4f](https://github.com/sensu/sensu-transport/commit/ddafe4fcee6e10281233623c013732aea2576568#diff-080a905f5ab39312d2a80c541ccb1583R125)
- [sensu-transport/commit/7bbf2437](https://github.com/nhooey/sensu-transport/commit/7bbf2437#diff-080a905f5ab39312d2a80c541ccb1583R184)

calebhailey:
@nhooey hi there! :wave: 

Thank you for submitting this issue. The behavior you're describing isn't a bug as much as the result of [changes to the Sensu Transport that defer processing until a connection is successfully establish][1]. I'm giving this issue the documentation label so we can clarify this change in behavior in the documentation. 

Having said that, we do agree that &ndash; because the API is used to provide visibility into Sensu's health via dashboards such as Uchiwa &ndash; the API should still respond to requests while it waits for transport connection to be established, so we'll label this as a feature as well, and work on getting it scheduled ASAP. 

/cc @amdprophet 

[1]: https://sensuapp.org/docs/latest/changelog

amdprophet:
Hi @nhooey,

[...]

PR #1298, opened 2016-05-30
API, check connections on request

portertech:
This PR allows the API to start listening immediately, potentially before it connects to Redis and the Sensu Transport. If the API receives a request before it connects to Redis and the Transport, it will return a 500 (error). If the API receives a request while it is reconnecting to either (or both) Redis or the Transport, it will return a 500 (error). `/info` and `/health` are excluded from reconnecting case.

Closes https://github.com/sensu/sensu/issues/1215

### DIFF
--- a/lib/sensu/api/process.rb
+++ b/lib/sensu/api/process.rb
@@ -315,7 +315,7 @@ module Sensu
         settings.cors.each do |header, value|
           headers["Access-Control-Allow-#{header}"] = value
         end
-        error! unless connected?
+        error! unless connected? || %w[/info /health].include?(env["REQUEST_PATH"])
         protected! unless env["REQUEST_METHOD"] == "OPTIONS"
       end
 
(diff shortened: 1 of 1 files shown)


amdprophet:
What do you think about including a more detailed error in the body? e.g. `{"error": "redis is not connected"}`? Dashboards could take advantage of more detailed errors and make troubleshooting easier.

portertech:
@amdprophet might as well :+1:

portertech:
`/info` and `/health` are now excluded from the connection check, providing necessary info for troubleshooting, beyond a string.

portertech:
```
{"error":"redis and transport connections not initialized"}
{"error":"not connected to redis"}
{"error":"not connected to transport"}
```

amdprophet:
:+1:

giphy

portertech:
horsehead

Issue #1333, opened 2016-06-13
No longer respecting REDACT on client. Passwords are shown in check


zbintliff:
With this commit:
https://github.com/sensu/sensu/commit/1917db8ff20a5ef6f40aa490f7a5658ae1253af5#diff-126205c86cd89c693bea304046d9006e

Since you now substitute on the check itself instead on a seperate `command` string. the check is modified. When you publish the check any passwords are now in raw text.  The redact is respected in the client hash but not the check page.

For example, before the check the check result would like so in uchiwa
```
command: /usr/bin/check-db.rb -h :::host|default::: -p :::password|default:::
```

After change it looks like:
```
command: /usr/bin/check-db.rb -h real_hostname -p real_password
```

this is a pretty big vuln where passwords need to be redacted.

zbintliff:
thanks for the fast response!  I saw this was merged into 0.25, was wondering if its worth putting in 0.24.2?

portertech:
@zbintliff 0.25 is a small (but important) release, https://github.com/sensu/sensu/issues?q=is%3Aissue+milestone%3A0.25+is%3Aclosed

PR #1335, opened 2016-06-13
Perform substitution on a clone of the check object

zbintliff:
Hope to fix issue laid out here: https://github.com/sensu/sensu/issues/1333

Because the substitution is done on the `check` object itself the command gets overwritten with the literal command that is ran.  This is then stored on redis and presented in Uchiwa. This is bad news for people that pass aws tokens or db passwords via the command line.  

I try to maintain same functionality that @UnitedMarsupials was aiming for while keeping it so that the un-substittued command and check is relayed up to the sensu-server.


zbintliff:
Closing because marshalling doesn't work and this is superceded by #1336

PR #1336, opened 2016-06-13
Fixed redaction + check client token substitution regression

portertech:
- Fixes https://github.com/sensu/sensu/issues/1333

### DIFF
--- a/lib/sensu/client/process.rb
+++ b/lib/sensu/client/process.rb
@@ -141,11 +141,12 @@ module Sensu
         @logger.debug("attempting to execute check command", :check => check)
         unless @checks_in_progress.include?(check[:name])
           @checks_in_progress << check[:name]
-          check, unmatched_tokens = object_substitute_tokens(check)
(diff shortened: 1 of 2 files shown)


zbintliff:
I was trying to fix this is #1335 as well. Aren't you still modifying the check object by passing it to the substitute function?

portertech:
@zbintliff no, initial object is duplicated, via `dup()`. Test verifies this claim. 0.25.0 will be released today :+1:

portertech:
@zbintliff I am mistaken, there is no `dup()`, but it doesn't look like we are running into a mutation issue.

portertech:
@zbintliff I am going to quickly patch to also use a `dup()` to cover our bases. We are only concerned about maintaining the behaviour of token subs in the check command, not the other attributes.

portertech:
https://github.com/sensu/sensu/pull/1337

zbintliff:
Sounds good. Thanks for swift response.

PR #1348, opened 2016-06-20
Use redis & transport callbacks with API

amdprophet:
Both `@redis` and `@transport` aren't guaranteed to be set by the time the API HTTP server starts serving requests. This PR implements the use of the `setup_redis` and `setup_transport` callbacks to ensure the connections are established.


cwjohnston:
Aw yeah blocks! 🎉

Issue #1235, opened 2016-04-16
Long check result sent to the client TCP socket fails to be processed


moises-silva:
When sending a large check result buffer to the client TCP socket the client fails to parse the first received chunk but MultiJson.load() does not throw any exception and instead returns a half-baked object of type Symbol which then causes an exception in process_check_result() when trying to access its elements.

I saw that recently Sean dropped MultiJson in favor of sensu-json. I have not tested if the same problem exists.

The json data we're seding is kind of sensitive (billing info) so I cannot paste the full log. Suffice to say is a json struct of about 45KB, but this is the extra logging I've added to the client socket.rb file to debug this problem:

```
{"timestamp":"2016-04-16T02:35:48.321689-0400","level":"warn","message":"Parsing data with MultiJson"}
{"timestamp":"2016-04-16T02:35:48.321985-0400","level":"warn","message":"Adapter","adapter":"MultiJson::Adapters::Yajl"}
{"timestamp":"2016-04-16T02:35:48.323086-0400","level":"warn","message":"Successfully parsed data with MultiJson"}
{"timestamp":"2016-04-16T02:35:48.323189-0400","level":"warn","message":"Watchdog is now cancelled"}
{"timestamp":"2016-04-16T02:35:48.323272-0400","level":"warn","message":"check object_id","id":"857308"}
{"timestamp":"2016-04-16T02:35:48.324687-0400","level":"warn","message":"check class","class":"Symbol"}
{"timestamp":"2016-04-16T02:35:48.324785-0400","level":"warn","message":"check to_s","class":"output"}
{"timestamp":"2016-04-16T02:35:48.324863-0400","level":"warn","message":"Setting status"}
{"timestamp":"2016-04-16T02:35:48.325019-0400","level":"warn","message":"failed to process check result from tcp socket, waiting for more data ...","error":"no implicit conversion of Symbol into Intege
r"}
```

All of those messages are printed in process_check_result(), so the check is supposed to be already parsed by MultiJSON, but the check var is of type 'Symbol' instead of Hash, so the next line throws an exception:

``` ruby
check[:status] ||= 0
```

The exception text is, as expected, saying "no implicit conversion of Symbol into Integer".

[...]

Issue #1304, opened 2016-05-31
Print running configuration to a temporary file


mattyjones:
I am looking for a language agnostic way to obtain the currently running config of a client. When using the new `-P` functionality it will compile the configuration that you pass into it. There should be a way to obtain the current configuration hash stored in the `settings` file.

There are a few ways to do this:
- When the client starts up dump it to /tmp/sensu_settings in machine readable json (perference)
- cmdline flag that will reach in and pull it out

portertech:
This will require a minor change to https://github.com/sensu/sensu-settings

portertech:
@mattyjones I am much more inclined to write the compiled/running configuration to a temp file :+1:

mattyjones:
👍

calebhailey:
The 0.25 release is going to hit today with fewer features than originally anticipated & primarily only some internal improvements. Moving this issue to 0.26.

moises-silva:
@mattyjones Could you check if PR #1352 is acceptable for your use case? you'd have to do this:

```bash
curl http://127.0.0.1:3030/settings > current-config.json
```

And you would get the client configuration as json

portertech:
I have renamed this issue to "Print running config to temp file", as a temporary file containing the running Sensu configuration can also be utilized by Sensu plugins (instead of using SENSU_LOADED_FILES).

portertech:
@mattyjones https://github.com/sensu/sensu/pull/1488 provides `/settings` with `&redacted=false`

PR #1352, opened 2016-06-28
Add http support to the client socket interface

moises-silva:
This is definitely not finished. I am looking for comments on whether you even want this functionality and if the implementation is going in the right direction or what to change, etc. Particularly be warned this is the first time I write more than a couple of lines of Ruby, so may be I am missing proper idioms etc; I'd appreciate feedback there too.

This PR started after our discussion on issue #1235 . Instead of adding a little dummy header with the length I decided it'd make more sense to simply use bare bones HTTP. This opens up also the opportunity to accept more commands other than "ping", such as "stats" perhaps for introspection. In general, it just felt easier to do HTTP that come up with my own header/value spec to specify the json length upfront, etc;

The old-style raw check notifications and ping command are still supported.

Finally, the HTTP_PARSER_AVAILABLE global was an easy way for me to introduce this on some systems, but I'd like your opinion on whether this should be configurable or if it's acceptable to make http-parser a fixed dependency.

Thanks!

### DIFF
--- a/lib/sensu/client/socket.rb
+++ b/lib/sensu/client/socket.rb
@@ -53,6 +53,22 @@ module Sensu
     # +WATCHDOG_DELAY+ (default is 500 msec) since the most recent chunk
     # of data was received, the agent will give up on the sender, and
     # instead respond +"invalid"+ and close the connection.
+    #
+    # == HTTP Protocol ==
+    #
+    # This is naturally implemented on top of the TCP protocol and
+    # requires sending an appropiate HTTP request. All requests
+    # will be responded with an appropiate HTTP response code and
+    # json body with a 'response' member typically set to 'ok' but
+    # that depends on the URL requested and whether the operation
+    # requested was successful or not.
+    # The following urls are available:
+    # /ping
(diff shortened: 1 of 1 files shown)


moises-silva:
@portertech I'll update the failed tests (and write new ones for the http thing) if you agree this http support thing makes sense (I hope so).

moises-silva:
Correct operation of some HTTP errors requires PR sensu/sensu-transport#33

moises-silva:
Note this PR also nows seems to address feature request #1304

portertech:
Hey @moises-silva, I created https://github.com/sensu/sensu/issues/1432 to reflect some of the ideas we discussed on our call as an alternative to this PR. The Sensu client TCP socket has already become a bit of a "Frankenstein" and we should avoid making it more of a monster :-)

portertech:
I completely agree that the Sensu client should use HTTP for an input :+1:

portertech:
This functionality was added to 0.27 by https://github.com/sensu/sensu/pull/1488

PR #1359, opened 2016-07-07
Include custom check attributes in check requests

portertech:
Fixes a customer reported issue, check definition client token substitution in custom attributes does not work with standard (pubsub) checks.


calebhailey:
LGTM 👍

calebhailey:
Fixes #1359

Issue #1328, opened 2016-06-10
NEW: Built-in Silencing 


calebhailey:
First class support for silencing is coming to Sensu Core! Silencing 2.0 will introduce a new, dedicated `/silence` API endpoint and specification. By building Sensu's new built-in silencing on Sensu subscriptions, Sensu will provide support for silencing events for an entire infrastructure, all results of specific checks, and/or a specific client (see: #1327). 

## `/silence (POST)` API

To update the silenced registry, post JSON data to the new `/silence` API.

For example, to silence events for clients with the subscription "load-balancer":

~~~ json 
{
  "subscription": "load-balancer",
  "expires": 1468881576,
  "reason": "scheduled maintenance",
  "creator": "portertech"
}
~~~

For example, to silence events for the check "haproxy-backends` when the client has the subscription "load-balancer":

~~~ json 
{
  "subscription": "load-balancer",
  "check": "haproxy-backends",
  "expires": 1468881576,
  "reason": "scheduled maintenance",
  "creator": "portertech"
}
~~~

For example, to silence events for the client "i-424242", leveraging its automatic per-client subscription https://github.com/sensu/sensu/issues/1327:

~~~ json 
{
  "subscription": "client:i-424242",
  "expires": 1468881576,
  "reason": "troubleshooting issue",
  "creator": "portertech"
}
~~~

For example, to silence events for the check "check-cron":

~~~ json 
{
  "check": "check-cron",
  "expires": 1468881576,
  "reason": "troubleshooting issue",
  "creator": "portertech"
}
~~~

## `/silenced (GET)` API 

Query the silenced registry: 

~~~ json
[
  {
    "subscription": "load-balancer",
    "check": "haproxy-backends",
    "expires": 1468881576,
    "reason": "scheduled maintenance",
    "creator": "portertech"
  },
  {
    "subscription": "client:i-424242",
    "expires": 1468881576,
    "reason": "troubleshooting issue",
    "creator": "portertech"
  },
  {
    "check": "check-cron",
    "expires": 1468881576,
    "reason": "troubleshooting issue",
    "creator": "portertech"
  }
]
~~~

## `/silenced/subscriptions/:subscription (GET)` API

Fetch a specific entry from the silenced registry, per subscription (e.g. `load-balancer`): 

~~~ shell
$ curl -s http://localhost:5678/silenced/subscriptions/load-balancer
[
  {
    "subscription": "load-balancer",
    "check": "haproxy-backends",
    "expires": 1468881576,
[...]

Issue #1327, opened 2016-06-10
Automatically create per-client subscriptions


calebhailey:
The Sensu client should automatically create a subscription called `client:client-name` (e.g. `client:i-424242`) for every client. This will allow for other features in Sensu to target specific clients or groups of clients using subscriptions.

calebhailey:
We'll need to coordinate w/ @palourde on this to make sure these subscriptions don't show up in the list of subscriptions for filtering views; h/t: https://github.com/sensu/sensu/issues/1316#issuecomment-224005599

troyready:
For transport methods that don't allow the client to connect to queue (e.g. the excellent [sensu-transport-snssqs](https://github.com/SimpleFinance/sensu-transport-snssqs)), would be nice to be able to optionally disable this or allow some other way of not [connecting to the transport for each subscription](https://github.com/sensu/sensu/blob/v0.25.6/lib/sensu/client/process.rb#L264).

cwjohnston:
Opened sensu/sensu-settings#40 to implement this feature.

@troyready I am not familiar with the SQS transport internals beyond my reading of the source code in [Sensu client](https://github.com/sensu/sensu/blob/master/lib/sensu/client/process.rb#L260-L282) and the [SQS transport](https://github.com/SimpleFinance/sensu-transport-snssqs/blob/master/lib/sensu/transport/snssqs.rb#L79-L103), but my interpretation of these two methods leads me to believe that this change will have no impact on SQS users.

Feel free to test this for yourself by adding a subscription to one or more of your clients, e.g. assuming client name i-424242

```json
{ "client": { "subscriptions": ["client:i-424242"] } }
```

cwjohnston:
@troyready actually as I dig deeper it seems to me that this `subscribe()` method defined in SQS transport is intended to be accessed only by the server process. SQS transport should probably add logic making `subscribe()` a no-op for clients.

troyready:
@cwjohnston what we're trying to do is avoid the [`@transport.subscribe`](https://github.com/sensu/sensu/blob/v0.25.6/lib/sensu/client/process.rb#L269) calls on sensu-clients. We can do that currently (at least as of 0.23, I haven't tested the newer releases but I think it's the same) by not defining any subscriptions on the clients.

Open to whatever the best way to handle it is.

portertech:
@troyready I believe that the SQS transport only needs to no-op subscribe() calls when the pipe type is fanout.

troyready:
[...]

PR #1364, opened 2016-07-08
Added standalone_mode, clients do not subscribe to the queue in this …

runningman84:
…case (important for snssqs transport)

### DIFF
--- a/lib/sensu/client/process.rb
+++ b/lib/sensu/client/process.rb
@@ -32,6 +32,7 @@ module Sensu
         @standalone_mode = @settings[:client][:standalone_mode] || false
         @checks_in_progress = []
         @sockets = []
+        @mode = 'client' 
       end
 
       # Create a Sensu client keepalive payload, to be sent over the
--- a/lib/sensu/client/process.rb
+++ b/lib/sensu/client/process.rb
@@ -7,6 +7,7 @@ module Sensu
       include Daemon
 
       attr_accessor :safe_mode
+      attr_accessor :standalone_mode
 
       # Create an instance of the Sensu client process, start the
       # client within the EventMachine event loop, and set up client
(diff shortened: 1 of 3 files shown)


runningman84:
@portertech can you give me a hint what is failing the tests? The client, server and api works fine in my machine with this modifications?

syepes:
@runningman84 Is adding this feature going to be compatible with https://github.com/sensu/sensu-transport/issues/36? I will be creating a pull in the following weeks

runningman84:
@syepes I do not see a reason why this feature would not be compatible. The standalone mode just ensure that the clients itself do not receive anything from the queue. This was important for us in order to use special transports like snssqs safely.

syepes:
@runningman84 Effectively it should still work if the client is still sending its data to the results queue.

runningman84:
@portertech I would really appreciate an answer.

runningman84:
@calebhailey maybe you can take a look?

portertech:
@runningman84 why does the transport need a mode? I can see the reasoning behind disabling client subscriptions.

portertech:
These changes would cause issues with several 0.26 features, such as https://github.com/sensu/sensu/issues/1327, which the upcoming built-in silencing feature depends on https://github.com/sensu/sensu/issues/1328

portertech:
Why is this "important for snssqs transport"?

runningman84:
The full explanation can be found here:
https://github.com/SimpleFinance/sensu-transport-snssqs/issues/5

A transport mode would be useful to do stuff like this
```
        case @settings[:mode]
        when 'client'
          raise("transport does not work without sns client, please specify publishing_sns_topic_arn") if @sns.nil?
        when 'server'
          raise("transport does not work without sqs client, please specify consuming_sqs_queue_url") if @sqs.nil?
        end
```

My understanding is that the upcoming changes in 0.26 would add a client:instance-id subscription to each instance. This is not a problem if the standalone mode would be included. Otherwise we would have problems because snssqs does only work if the client has empty subscriptions.

runningman84:
This is not needed anymore if the pull request in snssqs transport gets merged.

PR #1372, opened 2016-07-13
Fix API /results when the last client in the registry has no checks

### LINKED ISSUE #1356: api: /results returning empty array.  potential for partial results.
I found a condition under which the `/results` interface of the sensu-api can return an empty array even when there are plenty of results.

Short version:  If the last client in the clients list returned from redis has no checks associated with it,  you'll get an empty array.    Workaround:  delete that client. 

Medium version:   The nested iteration of redis calls in `get_results()` are async and the code doesn't properly synchronize that all the async requests complete.   (also true in the earlier sintra handler in at least 0.23)  The only reason it works at all is because the requests and responses are mostly serialized through the single redis connection.  So it happens that responses do come back in the order they were requested. 

long version: 

[the code](https://github.com/sensu/sensu/pull/1324#issuecomment-225255464)  first fetches a list of clients then iterates through this list of clients fetching a list of check names for those clients,  then iterates through those check names fetching the actual check results.    The problem is the redis calls it makes are asynchronous.  which is why the callbacks for the check results do this check `client_index == clients.length - 1 && check_index == checks.length - 1`    which I assume is intended to indicate that all the results have returned.  However it actually only indicates that the last request that was dispatched  (the one in which check_index and client_index met the condition) was called.   the reason this mostly works is because the requests to redis while async are actually serialized through the redis connection so the responses come back in the order the requests were made.   however if someone tried to parallize  with multiple redis connections this would break more with partial result sets.  because of that things are still mostly ok. 

So when does it break?  A second problem of the same type exists.   There is a [second point of return](https://github.com/sensu/sensu/blob/v0.25.4/lib/sensu/api/routes/results.rb#L48-L50)  when there is a client with no checks associated with it  the callback for the check names is just if  `client_index == clients.length - 1` (true for the last client in the list) then return the current results.   `results` at this point is still empty because only the client check names requests have returned from redis because of the serialization of request/response through the single redis connection none of the check results that were dispatched  will have returned yet.  Therefore if your last client has not check names associated with it.  you will get an empty results set. 

spent a lot of time figuring this out.  and I'm not sure which way to go next.  

- I could delete the client with no checks and move on. 
- try to fix the code.

I was thinking using an [eventmachine iterator](https://github.com/eventmachine/eventmachine/blob/master/lib/em/iterator.rb )    or simply hacking in our own count of dispatched and returned callbacks. 

but I wasn't really sure how to setup a breaking test for the issue, so I stopped there.  any guidance might spur me back into action.

hope this helps.

portertech:
This pull request fixes https://github.com/sensu/sensu/issues/1356, using a Redis PING call to serialize the HTTP response, ensuring the response content includes all of the results.

I believe that `/results` is the only route that requires a change like this.

### DIFF
--- a/lib/sensu/api/routes/results.rb
+++ b/lib/sensu/api/routes/results.rb
@@ -46,7 +46,9 @@ module Sensu
                       end
                     end
                   elsif client_index == clients.length - 1
-                    respond
+                    @redis.ping do
+                      respond
+                    end
                   end
                 end
               end
--- a/lib/sensu/daemon.rb
+++ b/lib/sensu/daemon.rb
@@ -9,7 +9,7 @@ gem "sensu-extension", "1.5.0"
 gem "sensu-extensions", "1.5.0"
 gem "sensu-transport", "6.0.0"
 gem "sensu-spawn", "2.2.0"
-gem "sensu-redis", "1.4.0"
+gem "sensu-redis", "1.5.0"
 
 require "time"
 require "uri"
--- a/sensu.gemspec
+++ b/sensu.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |s|
   s.add_dependency "sensu-extensions", "1.5.0"
   s.add_dependency "sensu-transport", "6.0.0"
   s.add_dependency "sensu-spawn", "2.2.0"
-  s.add_dependency "sensu-redis", "1.4.0"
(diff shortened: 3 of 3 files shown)


amdprophet:
👍

cwjohnston:
Serialize it! 👍

Issue #1286, opened 2016-05-25
Sensu Client exception on `standalone` check when `interval` omitted and `publish: false`


nhooey:
When I have a check defined with `standalone: true`, `publish: false`, and `interval` omitted, I get a stack trace when restarting `sensu-client`.

There is no issue when including an arbitrary `interval`, but I don't want the check scheduled standalone on the client, I merely want it to exist so the check will fail when there are no results posted within the `TTL`.

**Check definition:**

```json
{
   "checks": {
    "ansible-pull": {
        "command": "false",
        "handlers": [
            "all"
        ],
        "publish": false,
        "standalone": true,
        "timeout": 1200,
        "ttl": 1800
    }
}
```

(Note that `"command": "false"` refers to the Linux command that always fails, not a JSON boolean value).

**Stack trace when restarting `sensu-client`:**
```
{"timestamp":"2016-05-25T22:06:47.311233+0000","level":"debug","message":"scheduling standalone checks"}
error reading event: A JSON text must at least contain two octets!
/opt/sensu/embedded/lib/ruby/gems/2.3.0/gems/sensu-0.23.2/lib/sensu/client/process.rb:262:in `calculate_execution_splay': undefined method `*' for nil:NilClass (NoMethodError)
        from /opt/sensu/embedded/lib/ruby/gems/2.3.0/gems/sensu-0.23.2/lib/sensu/client/process.rb:279:in `block in schedule_checks'
        from /opt/sensu/embedded/lib/ruby/gems/2.3.0/gems/sensu-0.23.2/lib/sensu/client/process.rb:274:in `each'
        from /opt/sensu/embedded/lib/ruby/gems/2.3.0/gems/sensu-0.23.2/lib/sensu/client/process.rb:274:in `schedule_checks'
        from /opt/sensu/embedded/lib/ruby/gems/2.3.0/gems/sensu-0.23.2/lib/sensu/client/process.rb:300:in `setup_standalone'
        from /opt/sensu/embedded/lib/ruby/gems/2.3.0/gems/sensu-0.23.2/lib/sensu/client/process.rb:371:in `bootstrap'
        from /opt/sensu/embedded/lib/ruby/gems/2.3.0/gems/sensu-0.23.2/lib/sensu/client/process.rb:381:in `block in start'
        from /opt/sensu/embedded/lib/ruby/gems/2.3.0/gems/sensu-0.23.2/lib/sensu/daemon.rb:230:in `block in setup_transport'
[...]

PR #1287, opened 2016-05-25
Only schedule a `standalone` check if the `interval` is set.

nhooey:
Closes #1286.

PR #1384, opened 2016-07-20
Do not schedule standalone checks w/o an interval

portertech:
This pull requests brings the Sensu client check execution scheduler in-line with the behaviour of the Sensu server check request scheduler. The Sensu client will no longer attempt to schedule the execution of standalone checks if the definition has `"publish": false` and/or doesn't have a `"interval"`.

Closes: https://github.com/sensu/sensu/issues/1286

This is an alternative to https://github.com/sensu/sensu/pull/1287


calebhailey:
LGTM 👍

PR #1392, opened 2016-07-28
Client Metrics Caching

moises-silva:
This PR introduces metrics caching. In the event of losing network connection, the client will move to use standalone metric checks and cache the results locally. Due to our requirements, we cannot simply put all the checks in the node, in the future we may be able, but for now I am caching the check requests as well to know what to self-schedule when the network connection is lost. I have a TODO to invalidate that cache every N minutes for example so if we stop receiving a check request it'll be removed from the cache.

This PR depends on changes to the rabbitmq transport to enable message acknowledgements to detect when a keepalive was not acknowledged by the server and declare the connection down faster. I will be doing a separate PR for that in sensu/transport, but I am doing changes in other areas that I wanted to first get some feedback on the whole idea of caching and the direction we're taking.

There's a couple of commits related to a different PR here, so ignore changes in socket.rb, is just that our internal branch has multiple changes we've done and I did not have the chance yet to separate this code in a different branch. I just want initial feedback for this feature.

The use of sqlite might be controversial, but I wanted to keep things simple, avoiding a separate process (e.g relying on redis). I think you guys may have plans to require redis on the client at some point anyways? if that happens should be relatively easy to move to another db backend.

There's things that I plan to make configurable like the max amount of metrics to cache (currently unlimited) and the sync rate (as to not overwhelm the server if hundreds of nodes go offline and back online at the same time due to datacenter issues).

### DIFF
--- a/lib/sensu/client/process.rb
+++ b/lib/sensu/client/process.rb
@@ -1,5 +1,6 @@
 require "sensu/daemon"
 require "sensu/client/socket"
+require "sensu/client/store"
 
 module Sensu
   module Client
@@ -30,6 +31,18 @@ module Sensu
         @safe_mode = @settings[:client][:safe_mode] || false
         @checks_in_progress = []
         @sockets = []
+
+        # For metric caching support
+        # TODO: Cache limit?
+        @metrics_cache = @settings[:client][:metrics_cache] || false
+        # TODO: Persist this to disk so it can survive restarts or a client crash?
+        # FIXME: If a check is discovered here but fails to be scheduled
+        # by the server once, we should remove it from this hash as it
+        # is likely removed from our check list in the server
+        @metrics_checks_cache = {}
+        @metrics_cache_store = nil
+        @metrics_sync_timer = nil
+        @ack_count_on_keepalive = -1
       end
 
       # Create a Sensu client keepalive payload, to be sent over the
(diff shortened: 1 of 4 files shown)


moises-silva:
oh yeah I'm aware there's conflicts and that this isn't finished. This may not work well with self-scheduled checks, (or may be it does) but I have it on my plate to try it and fix it if needed. I am a new to ruby so help in that area would be appreciated as well (just learned the hard way that ruby treats ```if 0``` as true!).

portertech:
@moises-silva I wonder if it makes sense to introduce the concept of publish "pending" to the Sensu Transport API. This would also add publish pending (internal bounded queue) to things like the Sensu server Transport event handler. We could have the behaviour disabled by default, enable with something like `"transport": {"pending": true}`. WDYT?

portertech:
I would prefer to implement something more generic, with a wide and consistent application.

moises-silva:
@portertech Moving caching in the transport layer works for us. I'm unsure however if you mean every transport would be responsible for caching or adding a common layer in the base transport class? I'll update this PR anyways with our latest changes later this week. We already rolled this implementation into production but we'll be happy to rework this in another form for inclusion.

portertech:
@moises-silva the transport cache would have to be built into sensu-transport, so the underlying transports, e.g. rabbitmq, doesn't have to implement anything.

moises-silva:
That's what I thought, ok good.

Another thing, how would you handle the switch from server-scheduled to self-scheduled checks? That logic would have to stay in the client AFAICT?

maoe:
I'm [wondering](https://groups.google.com/d/msg/sensu-users/eV-78rfmaKQ/tRqwiHSWBAAJ) how to support stateful (e.g. edge-triggered) checks in Sensu. @moises-silva told me that it might be possible to support them on top of the store machinery in this PR.

@portertech If you're going to introduce some client-local state, it'd be great if it could also be used for stateful checks.

portertech:
We're not going to merge/implement this in Sensu at this time. The team at Sensu Inc is currently hard at work on the Sensu 2.0 alpha, which uses a ring buffer for all transport communication, which is capable of queuing data during loss of connectivity.

@moises-silva thank you for your hard work, I hope your fork with this functionality treats you well until we can share a Sensu binary :+1:

maoe:
@portertech Is there a plan to support stateful checks in Sensu 2.0?

Issue #1369, opened 2016-07-12
schedule a sensu check to run at specific times? sensu-cron...


joshzitting:
We have some checks that we only want to run at certain times off hours.. Like sunday morning at 1201 am. I would LOVE to see sensu be able to handle something like this..  It would be AWESOME to just be able to trigger these like a cron job but use the same monitoring/handler systems we already have... Any takers?? 
Thanks guys!!

joshzitting:
Sorry I meant for a standalone check. Hope we were on the same page! thanks guys! cant wait!!

joshzitting:
I think this is taken care of in #1397, and #1572 cant wait to test it!

PR #1397, opened 2016-07-30
Add cron to sensu

chriszacny:
Hi folks,

This is a relatively simple add of the "cron" attribute; it enables cron expressions that effectively override the "interval" attribute on a check. I've found this very useful for several of our batch checks that need to run at very specific times in house. Unless I'm missing a way to do this using interval.

I'd be happy to tidy this up as needed (for example, "interval" is still required on a check even if you specify "cron"); or simply request that this be added as a feature in a future release.

Thanks!

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -693,7 +693,8 @@ module Sensu
       # @param cron_expression [String] The cron expression you want to use.
       # @param check [Hash] definition.
       # @param create_check_request [block] reference to block to execute on next cron fire.
-      def schedule_cron(cron_expression, check, &create_check_request)
+      def schedule_cron(check, &create_check_request)
+        cron_expression = check[:cron]
         @logger.debug("Scheduling next cron tick for", :check => check)
         cron_parser = CronParser.new(cron_expression)
         now = Time.now
@@ -704,8 +705,22 @@ module Sensu
         @logger.debug("Time difference is: #{time_difference}")
         @timers[:leader] << EM::Timer.new(time_difference) do
           create_check_request.call
-          schedule_cron(cron_expression, check, &create_check_request)
+          schedule_cron(check, &create_check_request)
         end
+      end
+
+      # The block that creates the check request once the EM timer fires.
+      #
+      # @param check [Hash] definition.
+      def get_check_request_block(check)
+        create_check_request = Proc.new do
+            unless check_request_subdued?(check)
+              publish_check_request(check)
+            else
+              @logger.info("check request was subdued", :check => check)
+            end
+        end
+        return create_check_request
       end
 
       # Schedule check executions, using EventMachine periodic timers,
@@ -717,19 +732,9 @@ module Sensu
       # @param checks [Array] of definitions.
       def schedule_check_executions(checks)
         checks.each do |check|
-          
-          # This is an anonymous block of code, called in the block below
-          create_check_request = Proc.new do
-            unless check_request_subdued?(check)
-              publish_check_request(check)
-            else
-              @logger.info("check request was subdued", :check => check)
-            end
-          end
-
+          create_check_request = get_check_request_block(check)
           if check.has_key?(:cron)
-            cron_expression = check[:cron]
-            schedule_cron(cron_expression, check, &create_check_request)
+            schedule_cron(check, &create_check_request)
           else
             execution_splay = testing? ? 0 : calculate_check_execution_splay(check)
             interval = testing? ? 0.5 : check[:interval]
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -2,6 +2,7 @@ require "sensu/daemon"
 require "sensu/server/filter"
 require "sensu/server/mutate"
 require "sensu/server/handle"
+require "parse-cron"
 
 module Sensu
   module Server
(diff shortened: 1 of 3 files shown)


portertech:
Hi @chriszacny, this is neat!

I am curious to hear what other think about the use of the cron syntax. I wonder if there's even a comparable alternative.

This PR uses https://github.com/siebertm/parse-cron

portertech:
@cwjohnston @amdprophet WDYT? Implementation only requires a little cleanup, we would also need to apply these scheduler additions to the Sensu client for standalone.

cwjohnston:
@portertech @chriszacny I think this is an interesting idea, and one that has been requested elsewhere as well (see #1369). Is it the case that we can make this work for both pub-sub and standalone checks?

Although I think cron syntax isn't very intuitive, I believe it is sufficiently ubiquitous to make the case for adopting the same syntax for scheduling checks in Sensu.

chriszacny:
@portertech @cwjohnston Thanks guys. FYI @joshzitting as you had requested this feature.

I completely agree that cron syntax isn't intuitive; that said, it is the current standard. I'm totally cool with a cleaner version of it though using actual text that makes sense though. (dotbased notation? every_minute.every_hour.mon-fri.every_year).

The other thing to be careful of - as it stands, Sensu isn't (IMHO) a scheduler. It really is an event message receiver / mutator / forwarder. That's why I thought this (cron-based scheduling) would really just be a backup / override for interval... but that may cause confusion.

I think it would be relatively straightforward to implement in standalone checks (code looks identical to the server), but as it stands, it is only implemented for the server.

joshzitting:
I cant think of a reason to do this for subscriptions based checks. (but we only use standalone checks). As far as the syntax IDC. I am just fine with the normal cron or the dot-based. We will eventually use the chef cookbook to create these. So I am flexible. However we would like to use this on sensu clients as well as some sensu servers (which are also clients) so I think it makes more sense to have this functionality as part of the sensu-client than the server? Unless I am missing something? 
Thanks guys!

chriszacny:
Yeah @joshzitting that is a good point. Usually, if you were doing some large scale scheduled operation / check, it might make more sense to use something like chef/ansible to execute the operation. For smaller scale / atomic application-based checks, you'd probably run these from a single client, OR push directly from the application to the client TCP port to the server (which is what I prefer).

I'd still personally lean towards implementing for both server and client simply to reduce code inconsistency (looks like process.rb in sensu-client and sensu-server are somewhat similar; there might be an opportunity for refactoring there).

portertech:
I would like to see this functionality available in both check execution schedulers, pubsub (server) and standalone (client), perhaps the majority of the logic can be put in utilities.rb?

absolutejam:
Out of curiosity, is there any movement on this guys? I think this would be an amazing feature (and to me, it's the last big one missing for my uses) especially for subscription-based checks as it suddenly means I have a single, central way to scrape check info.

lib/sensu/server/process.rb
```diff
@@ -683,6 +684,45 @@ def calculate_check_execution_splay(check)
...
+      # check will still execute 'x' seconds from the original calculated 'fire' time
+      # even if that time has already passed. So watch out for that.
+      # 
+      # @param cron_expression [String] The cron expression you want to use.
+      # @param check [Hash] definition.
+      # @param create_check_request [block] reference to block to execute on next cron fire.
+      def schedule_cron(check, &create_check_request)
+        cron_expression = check[:cron]
+        @logger.debug("Scheduling next cron tick for", :check => check)
+        cron_parser = CronParser.new(cron_expression)
```
baweaver: You could just inline `cron_expression`:

```ruby
cron_parser = CronParser.new(check[:cron])
```

lib/sensu/server/process.rb
```diff
@@ -683,6 +684,45 @@ def calculate_check_execution_splay(check)
...
+      # The block that creates the check request once the EM timer fires.
+      #
+      # @param check [Hash] definition.
+      def get_check_request_block(check)
+        create_check_request = Proc.new do
+            unless check_request_subdued?(check)
+              publish_check_request(check)
+            else
+              @logger.info("check request was subdued", :check => check)
+            end
```
baweaver: Odd indentation, but nitpick

lib/sensu/server/process.rb
```diff
@@ -683,6 +684,45 @@ def calculate_check_execution_splay(check)
...
+        end
+        return create_check_request
```
baweaver: Explicit `return` isn't needed

spec/server/process_spec.rb
```diff
@@ -662,4 +662,15 @@
...
     end
   end
+
+  it "can schedule checks using cron" do
+    async_wrapper do
+      check = check_template
+      check[:cron] = "* * * * *"
+      create_check_request = @server.get_check_request_block(check)
+      @server.schedule_cron(check, &create_check_request)
+      expect(@server.instance_variable_get(:@timers).length == 1)
```
baweaver: You'll get better error messages with:

```ruby
expect(@server.instance_variable_get(:@timers).size).to eq(1)
```

Which'd return something like: "Expected expression to return 1, got 5" instead of "Expected expression to be true"

portertech:
@chriszacny thank you so much for this pull request and thank you for your patience. This pr has served as an fantastic feature request and implementation example. To get it working with the latest Sensu changes and apply it to the check standalone scheduler, I have created https://github.com/sensu/sensu/pull/1572, and updated sensu-settings to validate check `cron`.

@baweaver thank you for your input, I have tried to incorporate it into https://github.com/sensu/sensu/pull/1572 :+1:

PR #1400, opened 2016-08-03
Built-in Silencing

portertech:
A implementation of https://github.com/sensu/sensu/issues/1328

Closes: https://github.com/sensu/sensu/issues/1328

Example event data with silencing attributes:

``` json
{
  "check": {
    "name": "check-http",
    "...": "..."
  },
  "client": {
    "name": "i-424242",
    "subscriptions": [
      "webserver",
      "staging"
    ],
    "...": "..."
  },
  "silenced": true,
  "silenced_by": [
    "*:check-http",
    "staging:*"
    "client:i-424242:*"
  ]
}
```

### DIFF
--- a/lib/sensu/api/http_handler.rb
+++ b/lib/sensu/api/http_handler.rb
@@ -95,8 +95,8 @@ module Sensu
           data = Sensu::JSON.load(@http_content)
           valid = data.is_a?(Hash) && rules.all? do |key, rule|
             value = data[key]
-            (value.is_a?(rule[:type]) || (rule[:nil_ok] && value.nil?)) &&
-              (value.nil? || rule[:regex].nil?) ||
+            (Array(rule[:type]).any? {|type| value.is_a?(type)} ||
+              (rule[:nil_ok] && value.nil?)) && (value.nil? || rule[:regex].nil?) ||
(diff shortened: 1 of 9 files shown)


calebhailey:
This looks great! 

cantwait

cwjohnston:
Looks legit to me! 😻

portertech:
@cwjohnston @amdprophet thinking we should move the `event_silenced?()` call into to `process_event()`, reducing calls to Redis. Currently, every check result requires N number of Redis operations for silencing, where N is the # of client subscriptions * 2 (subscription and subscription + check) + 1 (just the check). Moving the `event_silenced?()` call to `process_event()` would only making these calls happen when an event meets conditions for handling.

portertech:
But, event bridges would not get silencing ... :-|

cwjohnston:
I think moving `event_silenced?()` check into `process_event()` is a good idea 👍

Not applying silencing to event bridges might be ok, at least in this initial implementation.

portertech:
Hmm, we do however need to update the event registry with silenced data :-1: So we cannot move it.

portertech:
Added a single conditional to `event_silenced?()` that greatly reduces the number of Redis calls, disabling silencing of OK events (except resolve).

amdprophet:
👍

PR #1406, opened 2016-08-08
Ensure /health API returns 204 No Content for "OK" responses

### LINKED ISSUE #1405: /health is returning incorrect HTTP status message
Hi,

On https://sensuapp.org/docs/0.25/api/health-and-info-api.html#health-get-specification `/health` is specified to return `204 (No Content)` however it has started returning the HTTP status `204 No Response`.

I believe this regression occurred in https://github.com/sensu/sensu/pull/1329/files#diff-38423adbb0ff4c5247b34ed86ea83dbcR254 as part of the 0.25 release.

cwjohnston:
As documented in #1405, the API is returning "204 No Response" instead of the
commonly accepted "204 No Content", which is also the documented behavior. This
change fixes the implementation and adds a test for the expected response
reason.

Closes #1405

### DIFF
--- a/lib/sensu/api/http_handler.rb
+++ b/lib/sensu/api/http_handler.rb
@@ -248,11 +248,11 @@ module Sensu
         respond
       end
 
-      # Respond to the HTTP request with a `204` (No Response)
+      # Respond to the HTTP request with a `204` (No Content)
       # response.
       def no_content!
         @response_status = 204
-        @response_status_string = "No Response"
+        @response_status_string = "No Content"
         respond
       end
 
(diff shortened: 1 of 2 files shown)


cwjohnston:
If there's any value in asserting the same http_reason expectation on all other 204s I can add that as well.

portertech:
:+1:

portertech:
@cwjohnston I am surprised that anyone cared about a status string :-)

cwjohnston:
I'm glad someone's keeping us honest :)

portertech:
@cwjohnston looks great, merging!

Issue #1416, opened 2016-08-17
Sensu API check routes return information on standalone checks


amdprophet:
## Expected Behavior
The `/checks` & `/check/:check` routes in Sensu API should only show subscription checks, not standalone checks.

## Current Behavior
The `/checks` & `/check/:check` routes in Sensu API will return information on both subscription checks and standalone checks.

## Possible Solution
Reject standalone checks prior to rendering the two API routes.

## Steps to Reproduce (for bugs)
1. Place both subscription checks and standalone checks on a node 
2. Place standalone checks on a non-API node
3. Hit the `/checks` route on the API. The results will only include standalone checks that exist on the server.

## Context
Standalone checks will be picked up by Sensu API if a Sensu Client exists on the same node and they both load checks from the same configuration file and/or directories. Including standalone checks that only exist on the node running Sensu API may be misleading to users as they may expect the API to also show information on standalone checks that exist outside of the API server.

## Your Environment
* Sensu version used: 0.25.7
* Operating System and version (e.g. Ubuntu 14.04): Arch
* Transport and runtime versions (e.g. RabbitMQ 3.6.2 on Erlang 18.2): RabbitMQ 3.6.5 on Erlang 1.19.0

PR #1417, opened 2016-08-17
Reject standalone checks in API before serving "/checks" routes

amdprophet:
## Description
The `/checks` & `/check/:check` routes in Sensu API should only show subscription checks, not standalone checks.

## Related Issue
https://github.com/sensu/sensu/issues/1416

## Motivation and Context
Standalone checks will be picked up by Sensu API if a Sensu Client exists on the same node and they both load checks from the same configuration file and/or directories. Including standalone checks that only exist on the node running Sensu API may be misleading to users as they may expect the API to also show information on standalone checks that exist outside of the API server.

## How Has This Been Tested?
Wrote test specs. Ensured they failed prior to the fix and that they passed after the fix.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/lib/sensu/api/routes/checks.rb
+++ b/lib/sensu/api/routes/checks.rb
@@ -7,14 +7,14 @@ module Sensu
 
         # GET /checks
         def get_checks
-          @response_content = @settings.checks
+          @response_content = @settings.checks.reject { |c| c[:standalone] }
           respond
         end
 
         # GET /checks/:check_name
         def get_check
           check_name = parse_uri(CHECK_URI).first
-          if @settings[:checks][check_name]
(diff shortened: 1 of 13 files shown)


lib/sensu/api/routes/checks.rb
```diff
@@ -7,14 +7,14 @@ module Checks
 
         # GET /checks
         def get_checks
-          @response_content = @settings.checks
+          @response_content = @settings.checks.reject { |c| c[:standalone] }
```
portertech: `c` -> `check`

portertech:
@amdprophet looks good, please target release/0.26

fessyfoo:
tx for this. 😄

cwjohnston:
Closes #1416

PR #1419, opened 2016-08-18
Added occurrences_watermark event attribute

portertech:


cwjohnston:
In the process of trying to implement an improvement over the existing occurrences filter, we have come to the following conclusions:

1. `last_ok` and `last_state_change` alone aren't sufficient to implement an objectively better filter 
2. `last_ok` doesn't have tests, also doesn't work as we expected
3. adding a "high water mark" for occurrences (`occurrences_watermark`) gives us the additional context needed for occurrence filtering to determine if an incident has exceeded check occurrences threshold, without requiring a comprehensive history.

I trust someone will correct me on any or all of these points if I've described this incorrectly. :)

portertech:
@cwjohnston your analysis is correct :+1:

fessyfoo:
I didn't like this at first,  thinking that it might be counter intuitive to base an occurrences filter off of and perhaps an `occurrences_total` (total of all events in this incident)  might lead to less confusion.

however when thinking about it,  I realize it will definitely work, in particular to allow the _always handle a resolve event  except in the one condition where I have filtered all previous events_

So,  nice idea 👍 

thinking about what I'd want a handler to do  when switching between `status: 1` and `status: 2`, red and yellow states, perhaps multiple times before eventually resolving hurts my head.    in some cases I want an `occurrences_watermark` and others I think i want an `occurrences_total`

I'm sure most people just don't want to have to think about it.

PR #1436, opened 2016-09-06
Handle SyntaxError exceptions from eval filters

cwjohnston:
## Description

When a filter contains a syntax error, the server currently raises an unhandled
`SyntaxError` exception, causing the daemon process to exit. This change explicitly
catches `SyntaxError` exceptions in addition to the `StandardError` exception
class which is caught by default when calling `rescue`.

## Related Issue
Closes sensu/sensu-enterprise#159

## Motivation and Context

I believe that ideally server should not crash due to configuration syntax errors once it has begun running. As we lack event context to  reasonably `eval` every filter at startup checking for syntax errors, I think it makes more sense to gracefully handle this exception and log the incidence.

## How Has This Been Tested?

Added a test to the filter spec to raise a `SyntaxError` exception and assert that 
it is handled with a false match on the filter.

## Types of changes

## Checklist:

### DIFF
--- a/lib/sensu/server/filter.rb
+++ b/lib/sensu/server/filter.rb
@@ -115,7 +115,7 @@ module Sensu
           begin
             value = Marshal.load(Marshal.dump(raw_value))
             !!Sandbox.eval(eval_string, value)
-          rescue => error
+          rescue StandardError, SyntaxError => error
             @logger.error("filter attribute eval error", {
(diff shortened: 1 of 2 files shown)


portertech:
:+1: 
94bf04ce-44e5-11e4-99ae-7fa5af8c9e8b

PR #1441, opened 2016-09-09
[server] ensure per-client subscription is added when creating a proxy client

### LINKED ISSUE #1439: Per-client silencing doesn't work for proxy clients
I created a silence for a check on a specific client, however I still then get notifications for that check/client combinations.

## Expected Behavior
The silence silenced notifications for the check and client.

## Current Behavior
` curl -s http://localhost:4567/silenced | jq`
```json
[
 {
    "id": "client:foo:check_chrony",
    "subscription": "client:foo",
    "check": "check_chrony",
    "reason": null,
    "creator": null,
    "expire_on_resolve": true,
    "expire": -1
  }
]
```
Yet I still get email alerts for UNKNOWN events.

## Your Environment
* Sensu version used: 0.26.1
* Operating System and version (e.g. Ubuntu 14.04): RHEL 7

cwjohnston:
## Description

In sensu/sensu-settings#40 we added per-client subscriptions which are a requirement for per-client native silencing added in 0.26, but this implementation doesn't cover proxy clients.

When processing events, the server calls `retrieve_client` to request client data from the registry using the client name. If a client with is not defined, we create a minimal entry in the client registry to represent the proxy client.

This change ensures that per-client subscriptions are added when a proxy client is registered.

## Related Issue

Closes #1439

## Motivation and Context

Fixes per-client silencing for proxy clients.

## How Has This Been Tested?

Added a unit test to ensure proxy clients are created with a per-client subscription.

## Types of changes

## Checklist:

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -575,7 +575,7 @@ module Sensu
         {
           :name => name,
           :address => "unknown",
-          :subscriptions => [],
+          :subscriptions => ["client:#{name}"],
           :keepalives => false,
           :version => VERSION
         }
--- a/spec/server/process_spec.rb
+++ b/spec/server/process_spec.rb
@@ -488,6 +488,7 @@ describe "Sensu::Server::Process" do
                   redis.hget("events:i-888888", "test") do |event_json|
                     event = Sensu::JSON.load(event_json)
                     expect(event[:client][:address]).to eq("unknown")
+                    expect(event[:client][:subscriptions]).to include("client:i-888888")
                     async_done
                   end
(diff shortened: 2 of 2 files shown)


cwjohnston:
Not sure what we should do to handle this for existing proxy clients... cc/ @portertech

portertech:
@cwjohnston this fix is sufficient :+1: Existing JIT clients will need to be deleted or updated (via POST /clients) to have the client sub required for client silencing. Any attempts to bridge the transition/upgrade would create a performance impact, requiring client inspection on registry update etc.

PR #1458, opened 2016-09-23
[HOTFIX] track in progress check extension executions

portertech:
We recently discovered that the Sensu client is not tracking in progress check extensions executions. This pull request implements the same logic used by check command execution. This pull request also implements check duration for check extension execution.


amdprophet:
Looks good!

PR #1475, opened 2016-10-20
Add Silenced API endpoint for retrieving entries by ID

### LINKED ISSUE #1465: Silencing API should support retrieving specific entries via ID
## Expected Behavior

As a Sensu operator, if I know the ID (subscription and check name, joined with a colon) of a silencing entry, I should be able to retrieve that specific entry by it's ID.

## Current Behavior

As reported by Alex Rodrigues via sensu-users:

> When using the new built-in silencing, i seem to be only able to query for a specific client (e.g. curl -u admin http://sensu-api:4567/silenced/subscriptions/client:client-name) or a specific check (e.g. curl -u admin http://sensu-api:4567/silenced/checks/check-name) but it does not “seem” to support a combination like we did with stashes.

## Possible Solution

Implement a route for /silenced/id/:id or similar, which would answer HTTP GET to return a single silencing entry as JSON hash, e.g.:

* `client:i-424242` being the per-client subscription for a specific Sensu client
* `check_ntpd` being the name of a check run by that client

I use POST to create an entry
```
$ curl -s -i -X POST \
-H 'Content-Type: application/json' \
-d '{"subscription": "client:i-424242", "check": "check_ntpd"}' \
http://localhost:4567/silenced

HTTP/1.1 201 Created
```

Under proposed /silenced/id/:id route I would be able to retrieve a specific check by its ID:

```
curl http://localhost:4567/silenced/id/client:i-424242:check_ntpd

{
  "expire": -1,
  "expire_on_resolve": false,
  "creator": null,
  "reason": null,
  "check": "check_ntpd",
  "subscription": "client:i-424242",
 "id": "client:i-424242:check_ntpd"
}
```

## Context

Although it is possible to find a specific override by ID using the list of all silencing overrides returned by `GET /silenced`, this is not optimal for many use cases.

## Your Environment

* Sensu version used: 0.26.4
* Operating System and version (e.g. Ubuntu 14.04): Ubuntu 14.04
* Transport and runtime versions (e.g. RabbitMQ 3.6.2 on Erlang 18.2): redis transport

cwjohnston:
## Description

Adds a new endpoint to the [Silenced API]( https://sensuapp.org/docs/latest/reference/silencing.html) for retrieving entries by their ID.

## Related Issue

Closes #1465

## Motivation and Context

As described in #1465, sometimes you just want to retrieve a silencing entry by its known ID, or know if a specific entry exists.

## How Has This Been Tested?

Added unit tests in api process spec.

## Types of changes

## Checklist:

### DIFF
--- a/lib/sensu/api/routes/silenced.rb
+++ b/lib/sensu/api/routes/silenced.rb
@@ -46,7 +46,7 @@ module Sensu
         # POST /silenced
         def post_silenced
           rules = {
-            :subscription => {:type => String, :nil_ok => true},
+            :subscription => {:type => String, :nil_ok => true, :regex => /\A[\w\.\-\:]+\z/},
             :check => {:type => String, :nil_ok => true, :regex => /\A[\w\.-]+\z/},
             :expire => {:type => Integer, :nil_ok => true},
             :reason => {:type => String, :nil_ok => true},
--- a/lib/sensu/api/routes/silenced.rb
+++ b/lib/sensu/api/routes/silenced.rb
@@ -3,7 +3,7 @@ module Sensu
     module Routes
       module Silenced
         SILENCED_URI = /^\/silenced$/
-        SILENCED_ID_URI = /^\/silenced\/ids\/([\w\.\-\*]+:[\w\.\-\*]+)$/
+        SILENCED_ID_URI = /^\/silenced\/ids\/([\w\.\-\*\:]+)$/
         SILENCED_SUBSCRIPTION_URI = /^\/silenced\/subscriptions\/([\w\.\-:]+)$/
         SILENCED_CHECK_URI = /^\/silenced\/checks\/([\w\.\-]+)$/
         SILENCED_CLEAR_URI = /^\/silenced\/clear$/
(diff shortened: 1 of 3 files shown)


spec/api/process_spec.rb
```diff
@@ -1528,6 +1528,78 @@
...
+            expect(http.response_header.status).to eq(200)
+            expect(body).to be_kind_of(Hash)
+            expect(body[:id]).to eq("test:*")
+            async_done
+          end
+        end
+      end
+    end
+
+    it "can retrieve entry for silencing all checks on all subscriptions" do
```
cwjohnston: Depending on perspective the bug might be that `:check => '*'` should be valid, and that `*:*`should silence everything, or that `:subscription => '*'` shouldn't be valid and silencing every check on every subscription is out of scope.

lib/sensu/api/routes/silenced.rb
```diff
@@ -3,6 +3,7 @@ module API
     module Routes
       module Silenced
         SILENCED_URI = /^\/silenced$/
+        SILENCED_ID_URI = /^\/silenced\/ids\/([\w\.\-\*]+:[\w\.\-\*]+)$/
```
portertech: This feels strange, as the route requires ':' in the query string or the generic 404 will be used. It still accomplishes the same functionality, as accepting any string as an ID will just push 404 responsibilities to the route method.

spec/api/process_spec.rb
cwjohnston: commit cc7ceb0 adds a regexp which excludes using the `*` wildcard as the value of  `subscription` attribute when creating a silencing entry.

portertech:
:+1:

PR #1482, opened 2016-11-02
IPv4/IPv6 dual-stack support

amdprophet:
This enables Sensu to have multiple bind addresses for local client sockets, which now listen on IPv4 & IPv6 by default. It also modifies the default API configuration to listen on both IPv4 and IPv6 by default.

In addition to the current changes, support for multiple bind addresses within the API should be considered.


amdprophet:
Closing this for now as this does not currently work on some platforms.

PR #1483, opened 2016-11-03
Use Sensu Transport 7.0.2 for DNS resolver

portertech:
This pull request bumps sensu-transport to 7.0.2 and fixes a client deregistration bug.

### DIFF
--- a/lib/sensu/client/process.rb
+++ b/lib/sensu/client/process.rb
@@ -496,8 +496,11 @@ module Sensu
       # deregistration check result if configured to do so.
       def stop
         @logger.warn("stopping")
+        last_state = @state
         pause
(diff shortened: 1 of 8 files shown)


lib/sensu/client/process.rb
```diff
@@ -496,8 +496,11 @@ def resume
       # deregistration check result if configured to do so.
       def stop
         @logger.warn("stopping")
+        last_state = @state
         pause
-        deregister if @settings[:client][:deregister] == true
+        if @settings[:client][:deregister] == true && last_state != :initializing
```
portertech: This fixes a bug where deregistration would trigger even when the client has not yet connected to the transport.

cwjohnston:
I like it!

Issue #1504, opened 2016-11-23
Sensu-Server crashes on `split': invalid byte sequence in UTF-8


mehmood-007:
## Current Behavior

Sensu-Server Crashes 

## Possible Solution

 "redis": {
    "password": "xxxxxxx",
    "master": "xxxxx",
    "sentinels": [
      {
        "host": "10.80.2.1",
        "port": 26379,
        "auto_reconnect": true,
        "reconnect_on_error": true
      },
      {
        "host": "10.80.2.2",
        "port": 26379,
        "auto_reconnect": true,
        "reconnect_on_error": true
      },
      {
        "host": "10.80.2.3",
        "port": 26379,
        "auto_reconnect": true,
        "reconnect_on_error": true
      }
    ],
    "auto_reconnect": true,
    "reconnect_on_error": true
  },

## Context

#Logs:
{"timestamp":"2016-11-23T02:54:00.693425+0100","level":"warn","message":"reconnecting to redis"}
#Backtrace
/opt/sensu/embedded/lib/ruby/gems/2.3.0/gems/sensu-0.26.3/lib/sensu/server/process.rb:322:in `split': invalid byte sequence in UTF-8 (ArgumentError)
        from /opt/sensu/embedded/lib/ruby/gems/2.3.0/gems/sensu-0.26.3/lib/sensu/server/process.rb:322:in `truncate_check_output'
        from /opt/sensu/embedded/lib/ruby/gems/2.3.0/gems/sensu-0.26.3/lib/sensu/server/process.rb:348:in `store_check_result'
        from /opt/sensu/embedded/lib/ruby/gems/2.3.0/gems/sensu-0.26.3/lib/sensu/server/process.rb:659:in `block in process_check_result'
        from /opt/sensu/embedded/lib/ruby/gems/2.3.0/gems/sensu-0.26.3/lib/sensu/server/process.rb:612:in `block in retrieve_client'
        from /opt/sensu/embedded/lib/ruby/gems/2.3.0/gems/sensu-redis-1.6.0/lib/sensu/redis/client.rb:315:in `dispatch_response'
        from /opt/sensu/embedded/lib/ruby/gems/2.3.0/gems/sensu-redis-1.6.0/lib/sensu/redis/client.rb:337:in `parse_response_line'
        from /opt/sensu/embedded/lib/ruby/gems/2.3.0/gems/sensu-redis-1.6.0/lib/sensu/redis/client.rb:365:in `receive_data'
[...]

PR #1505, opened 2016-11-23
Fix metric output truncate output encoding

portertech:
Force encoding to UTF-8 before split


portertech:
Closes https://github.com/sensu/sensu/issues/1504

cwjohnston:
Is there a test we can add for this without too much hassle?

portertech:
@cwjohnston yes, adding :+1:

Issue #1302, opened 2016-05-31
TTL check has a different interval from the original check


lp0:
If I have a check running every 5 minutes and the TTL expires on the last result, sensu starts calling the handler every 30 seconds but the check will not recover until the next 5 minute interval.

The message to the handler also increments the number of occurrences every 30 seconds which makes it impossible for the handler to identify the duration of that state or correctly determine the duration when the check is resolved.

Delaying an alert based on the number of occurrences fails to work properly because the number of occurrences goes up rapidly.

calebhailey:
@lp0 thank you for submitting this issue! This is an interesting problem indeed. There are a few other TTL-related issues that we're taking a look at in the upcoming 0.26 release (e.g. #1282). 

@portertech perhaps we should have the Sensu server compare the `ttl` and `interval` values during [`determine_stale_check_results`](https://github.com/sensu/sensu/blob/master/lib/sensu/server/process.rb#L862-L891)? I'm not sure if we store enough state (i.e. the time of the last TTL event) in order to be able to compare it against the check `interval` currently, but that doesn't seem too unreasonable to me.

portertech:
Without a 30 second TTL event update, the time since last check result information will be incorrect/outdated.

CVTJNII:
Suggest altering the interval in the message so that the handlers, which by default use refresh and interval to calculate the count to alert at, will automatically update for the interval change.  This should mitigate the problem when using the sensu-plugins handler, though it's a bit of a hack.

Ref: https://github.com/sensu-plugins/sensu-plugin/blob/master/lib/sensu-handler.rb#L171

portertech:
The check TTL override is available in Sensu >= 0.27.0 :+1:

PR #1514, opened 2016-12-08
Override check interval in TTL alarm with interval of TTL timer

CVTJNII:
…so that handler calculations are correct

Proposed partial solution for #1302: Extend the TTL check result to set the TTL timer interval in the check.  When using [Sensu-Plugin handlers](https://github.com/sensu-plugins/sensu-plugin/blob/ead9fd0a81ff3f099a0032f90ae6976d63a345c5/lib/sensu-handler.rb#L171) this will allow the handler to update the re-alert occurrence count if the TTL interval is not equal to the actual check interval.

Resolves the concern about re-alerts in #1302.  Does not address other concerns about the interval time changing.

Details of change:
- Turns the magic "30" value into a method argument default so that is can be passed around easier and be a little less magical
- Sets the :interval key in the TTL check result to the TTL timer interval value
- Adds test coverage

## Motivation and Context
If a check is set up to be slow, say every hour with a 6 hour realert, then the handler refresh occurrence value is 6.  If that check goes in to TTL alert with a 30 second interval, but still passing 1 hour interval to the handler from the underlying check, then the refresh goes from 1 hour to three minutes due to 30sec * 6 = 3 min.  This change mitigates this and ensures the check payload interval will match the rate from the TTL processor.

## How Has This Been Tested?
- Spec tests updated and tested with 'bundle exec rake'
- Tested in a lab environment by running the Sensu stack (server/api installed via gem specific_install into a container) and verified the interval changed from the check interval to the TTL interval (30).

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [X] I have read the **CONTRIBUTING** document.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

NOTE: (Re: Contributing) There hasn't been any discussion of my proposed solution in #1302 yet, but this change is simple enough that I believe it to be safe.


portertech:
:+1:

Issue #1512, opened 2016-12-06
event not triggering resolve action from handler while expire_on_resolve = true


cabecada:
sensu alerts are triggering handler events when action == create.
if the alert is silenced with expire_on_resolve, and alert is resolved, the event gets filtered even with 
expire_on_resolve.

```
#mon_check.json
#handlers are https://github.com/Yelp/sensu_handlers

{
  "checks": {
    "mon_check": {
      "command": "ls /tmp/mon_check",
      "dependencies": [

      ],
      "handlers": [
        "default"
      ],
      "interval": 10,
      "standalone": true,
      "timeout": 10,
      "subdue": {
        "days": {
          "all": [
            {
              "begin": "00:00",
              "end": "00:00"
            }
          ]
        }
      },
      "runbook": "",
      "team": "fake_ops",
      "alert_after": 10,
      "realert_every": "-1",
      "page": true,
      "ticket": false,
    }
  }
}
```

trigger an alert by 
```
touch /tmp/mon_check
sleep 3
rm /tmp/mon_check
```

it rightly triggers the "action=create" event.

now, silence the alert with expire on resolve(from uchiwa)
and verify alert has expire_on_resolve in its silenced data.

```
localhost:6379> mget "silence:client:foobar.com:mon_check"
1) "{\"id\":\"client:foobar.com:mon_check\",\"subscription\":\"client:foobar.com\",\"check\":\"mon_check\",\"reason\":\"pagerduty webhook\",\"creator\":null,\"expire_on_resolve\":true}"
 ```

the alert is silenced and the handler action is now filtered, all good till here.

now
```
touch /tmp/mon_check
```

the check is resolved, the redis silenced entry is removed, but the alert does not fire resolve event although alerts gets resolved.
this is painful, coz the new alerts get filtered as earlier alert is still ack'ed in pagerduty(handler).

* Sensu version used: 0.26.4
* Operating System and version (e.g. Ubuntu 14.04):
* Transport and runtime versions (e.g. RabbitMQ 3.5.2 on Erlang R16B03:

cabecada:
and this is only with expire_on_resolve.
handlers work fine (create and resolve trigger) without expire_on_resolve.

portertech:
Hi @cabecada, thank you for reporting this. We think that the behaviour that you desire is appropriate, I have added this to the 0.28 milestone.

cabecada:
[...]

PR #1519, opened 2016-12-14
Silenced resolution events with expire_on_resolve should be handled

portertech:
Closes https://github.com/sensu/sensu/issues/1512


portertech:
Hmm, we need to just exclude the silenced entry that expire on resolve, there may be others that do not.

Issue #1278, opened 2016-05-23
Question: Configuration for 50+ switches?


dR3b:
Hi

I have a default "sensu + uchiwa" installation running. The sensu-server and the
sensu-client are on the same VM.

I also added a simple "ping" check for 1 switch:

```
{
  "checks": {
    "sw1": {
      "command": "/opt/sensu/embedded/bin/check-ping.rb -h sw1",
      "subscribers": [
        "sw"
      ],
      "interval": 30
    }
  }
}
```

This is working with no problem. Now i want to add 50 more switches. Do I really have to repeate this step for every switch?

Thanks

kobbe:
Since commands can be any command it is quite easy to just write/script your own checks. Make for example a python/ruby/bash script that loops through all desired hosts.

I usually do it like this in linux:

```#!/bin/bash

#This program returns the greatest exit status

EXIT_STATUS=0
echo -n "Description of what the output is: "; COMMAND || EXIT_STATUS=$(($EXIT_STATUS>$? ? $EXIT_STATUS : $?))
echo -n "Description of what the output 2 is: "; COMMAND2 || EXIT_STATUS=$(($EXIT_STATUS>$? ? $EXIT_STATUS : $?))
exit $EXIT_STATUS
...
```

If you don't want to manually add all rows you can just make a loop.

dR3b:
Sorry, but that sounds terrible for me. That's not a solution that I expect for a "Monitoring for today's infrastructure" tool.

I need a "virtual" host for every switch. I think something like this:
```
{
  "checks": {
    "::SWITCHNAME::": ["sw1", "sw2", "sw3", "swN"],
      "source": "::SWITCHNAME::",
      "command": "/opt/sensu/embedded/bin/check-ping.rb -h ::SWITCHNAME::",
      "subscribers": ["sw"],
      "interval": 30
    }
  }
}
```
One check definition for all switches (sw1, sw2, swN).

yanc0:
There is no concept of distant check in sensu as far I know. All machines that you want to be monitored have to run the `sensu-client` daemon. It seems that sensu is not the right tool for network assets monitoring. I figured out distant monitoring with a little HTTP Client As a Service that I rapidly developed, It is called [pingmeback](https://github.com/yanc0/pingmeback) with a sensu plugin named [sensu-plugins-pingmeback](https://github.com/yanc0/sensu-plugins-pingmeback) for distant monitoring purposes. Maybe this can be a begin of an answer.

[...]

Issue #1259, opened 2016-05-06
Feature Request:  sources attribute for proxy clients


Craigeous:
I've been working with the proxy client and the source attribute in checks.  I am happy they exist, but I'm finding I have to create a check for each proxy client, which is fine, I just wish they provided a little more function.  

I'd like to be able to aggregate the checks as well, which you cannot do(from my knowledge) in the current form.  I was thinking, it would be nice if we had a sources attribute, that had an array of proxy clients, similar to handlers?  Below is an example I was thinking of to illustrate my idea.  Not sure if this is possible or not.

**Register clients:**
`curl -s -i -X POST -H 'Content-Type: application/json' -d '{"name":"switch-x","address":"0.0.0.0","environment": "infrastructure", "subscriptions":[],"keepalives":false,"}' http://sensu-server:4567/clients`

_same for switch-y, firewall-x, firewall-y, router-x, and router-y, etc..._

**Create the checks:**

```
{ 
  "checks": { 
    "check_ping_switch": { 
      "handlers": ["default", "mailer"],
      "command": "check-ping.rb -h :::source.address:::",
      "sources": ["switch-x", "switch-y"],
      "subscribers": ["infrastructure"],
      "interval": 30,
      "aggregate": true 
    },
    "check_ping_firewall": { 
      "handlers": ["default", "mailer"],
      "command": "check-ping.rb -h :::source.address:::",
      "sources": ["firewall-x", "firewall-y"],
      "subscribers": ["infrastructure"],
      "interval": 30,
      "aggregate": true 
    },
    "check_ping_router": { 
      "handlers": ["default", "mailer"],
      "command": "check-ping.rb -h :::source.address:::",
      "sources": ["router-x", "router-y"],
      "subscribers": ["infrastructure"],
      "interval": 30,
      "aggregate": true 
    } 
  } 
} 
```

When the check runs it will look at each entry in source and the address attribute of that client and insert that into the host.  You would have one check/aggregate for each type rather than one for each proxy client, but you'd still have individual clients.

lp0:
I would need :::source.name::: as well as as support for multiple source address families, e.g. a check for SSL using the IPv4 address of the source and a check for SSL using the IPv6 address of the source. Both of these would also require the name of the source to be able to verify the hostname on the certificate.

Samgarr:
+1

cwjohnston:
@Craigeous @lp0  thanks for opening this issue and describing your use cases. I'm tagging this for 
[...]

Issue #1371, opened 2016-07-13
Add ability to define proxy clients in the server configuration


Gillingham:
I have a fairly fixed set of proxy clients to represent a service that can be run on several servers in a cluster. It would be nice if I could statically define the client attributes for those proxy clients in a config file on the sensu server.

Right now the only way to do this to execute serveral HTTP API calls, and make sure they gets run every time the sensu-server or api gets restarted since that can and does sometimes lose the client data. Having to do that especially after some cases where sensu may crash or the server gets rebooted is far from ideal, thus my request for the ability to declare them in a static config file.

sstarcher:
The possibility to define the JIT source configuration statically would simplify some of my configuration.   Currently I define alerting at the JIT check level, but we have a static set of JIT sources that would simplify our setup.

cwjohnston:
In #1555 we're proposing to implement "proxy check requests", a new feature allowing a pubsub check definition to generate a unique check request for each client matching a defined set of client attributes.

Please have a look and share your thoughts?

cc: @Gillingham @sstarcher

portertech:
[Sensu 2.x](https://github.com/sensu/sensu-go) supports creating Entities via the Backend API 🎉  See https://docs.sensu.io/sensu-core/2.0/reference/entities/#how-do-entities-work to learn more about Entities!

PR #1555, opened 2017-02-13
Proxy check requests

portertech:
Publish a check request to the configured `subscribers` (e.g. `["round-robin:snmp_pollers"]`) for every Sensu client in the registry that matches the configured client attributes in `client_attributes` on the configured `interval` (e.g. `60`). Client tokens in the check definition (e.g. `"check-snmp-if.rb -h :::address::: -i eth0"`) are substituted prior to publishing the check request. The check request check `source` is set to the client `name`.

``` json
{
  "checks": {
    "foobar": {
      "command": "check-snmp-if.rb -h :::address::: -i eth0",
      "subscribers": [
        "round-robin:snmp_pollers"
      ],
      "interval": 60,
      "proxy_requests": {
        "client_attributes": {
          "keepalive": false,
          "device_type": "router",
          "device_manufacturer": "arista"
        }
      }
    }
  }
}
```


cwjohnston:
@portertech will the `request_generator` support attributes other than those in `client` scope? When I compare your example with what we do in filter definitions, it seems more consistent to nest `client` under `attributes` like this:

``` json
{
  "checks": {
    "foobar": {
      ...
      "request_generator": {
        "attributes": {
          "client": {
            "keepalive": false,
            "device_type": "router",
            "device_manufacturer": "arista"
          }
        }
      }
    }
  }
}
```

portertech:
@cwjohnston the feature is limited to the client registry and therefore only the client definition scope.

portertech:
Now to work out naming things and configuration scope.

cwjohnston:
Closes #1259 
Closes #1278

exeral:
from what I undertsand,
this check config
```
      "proxy_requests": {
        "client_attributes": {
          "keepalive": false,
          "device_type": "router",
          "device_manufacturer": "arista"
        }
```
means that it matches all clients like:
```
 "client": {
    "address": "10.1.2.3",
    "name": "switch1",
    "device_type": "router",
    "device_manufacturer": "arista"
    "keepalive": {
    },
    "subscriptions": [
      "default",
      "virtual",
      "another-subscription"
    ]
  }
```
```
 "client": {
    "address": "10.1.2.4",
    "name": "switch2",
    "device_type": "router",
    "device_manufacturer": "arista"
    "keepalive": {
    },
    "subscriptions": [
      "default",
      "another-subscription"
    ]
  }
```
???

so the snmp_poller will do
`check-snmp-if.rb -h 10.1.2.3 -i eth0`
`check-snmp-if.rb -h 10.1.2.4 -i eth0`

I'm right ?

cwjohnston:
@exeral you have the right idea 👍 

Note that the clients would also need `"keepalive": false` instead of `"keepalive": {}` in order to match.

exeral:
great ! that's what I was looking for

x70b1:
Very nice to see a progress on this topic.
exeral show us already an example. But does it work with subscriptions also?

Like this? I define a client and add a subscription.

```
curl -s -i -X POST -H 'Content-Type: application/json' -d '{"name": "web.server", "address": "12.23.34.45", "subscriptions":["webservers"]}' http://127.0.0.1:123/clients

```
{
  "checks": {
    "snmp": {
      "command": "check-snmp-if.rb -h :::address::: -i eth0",
      "subscribers": [
        "webserver"
      ],
      "interval": 60,
      "proxy_requests": {
        "client_attributes": {
          "keepalive": false
        }
      }
    }
  }
}
```

Gillingham:
This would be much more useful with #1371 supported as well, having to write a bunch of API requests every time the sensu data clears out for things like static clients is a bit of a pain vs just defining them in a configuration file.

dreamerm99:
Hi, I was wondering if there is a more detailed example on how to configure configs for external proxy client?  I see the example you had in documentation that utilizes client_attributes to filter with clients with the matched tags, and then executes the check against those client's address with the toke :::address:::.  


Where would I define configuration for external clients so the check config will be able to pull its address to perform checks?

xblitz:
any updates on the adding of defining static clients for use with the Proxy checks requests?
Since our sensu setup is installed via chef and everything is automated, it would be great if there was a simple way of defining all of the static client without doing API Calls

exeral:
Hi, I have the same problem. I'm wondering how to monitor dozen of snmp devices. I don't understand how/where they should be created (especially via puppet)
every doc talks about proxy client, but can I create proxy client through puppet ?

PR #1644, opened 2017-03-23
Sensu CLI defaults for config file & directory

portertech:
## Description
This pull-request introduces CLI option default values for `config_file` and `config_dirs`. The default values, `/etc/sensu/config.json` and `/etc/sensu/conf.d`, are only set if the file/directory resource exists.

These changes will help Sensu users run `--print_config` and `--validate_config` successfully in production, as many users forget to also use the `-c` and `-d` arguments.

## Types of changes

## Checklist:

### DIFF
--- a/spec/cli_spec.rb
+++ b/spec/cli_spec.rb
@@ -6,15 +6,18 @@ describe "Sensu::CLI" do
   include Helpers
 
   it "can provide default configuration options" do
+    expect(File).to receive(:exist?) do |value|
+      true
+    end
+    expect(Dir).to receive(:exist?) do |value|
+      true
+    end
     options = Sensu::CLI.read
-    if File.exist?("/etc/sensu/config.json")
-      expect(options[:config_file]).to eq("/etc/sensu/config.json")
-    end
-    if Dir.exists?("/etc/sensu/conf.d")
-      expect(options[:config_dirs]).to eq("/etc/sensu/conf.d")
-    end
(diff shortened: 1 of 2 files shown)


spec/cli_spec.rb
```diff
@@ -5,8 +5,16 @@
 describe "Sensu::CLI" do
   include Helpers
 
-  it "does not provide default configuration options" do
-    expect(Sensu::CLI.read).to eq(Hash.new)
+  it "can provide default configuration options" do
+    options = Sensu::CLI.read
+    if File.exist?("/etc/sensu/config.json")
```
cwjohnston: Should the specs stub `File.exist?` so the specs will actually be exercised?

portertech: @cwjohnston perhaps, conditional defaults are gross.

portertech:
@cwjohnston @amdprophet we need to make the call, :+1: or :-1: 

This does help users on Linux, run `--print_config` and `--validate_config`. Windows & BSD users will still need to provide `-c` and `-d`. We still must document using all of the arguments, regardless.

cwjohnston:
I'm in favor 👍  with the expectation that our service management scripts on all platforms will continue to explicitly define values for these flags.

amdprophet:
I'm mostly 👍  on this change. It would be nice to support default values for Windows & BSD but perhaps that can come later.

PR #1696, opened 2017-06-22
Implement /incidents as an alias for /events API route

### LINKED ISSUE #1690: API should help elucidate the distinction between incidents and events
## Expected Behavior

It should be easy to understand the distinction between events and incidents.

## Current Behavior

As someone who aims to help folks become proficient with Sensu, I find that "event" is an overloaded term.

In event processing terms, each check result becomes an event when it is correlated with information from the client registry and other additional context. Subsequently, each event is  evaluated to determine whether or not it should be sent to a handler. 

In API terms, the events returned via the `/events` endpoint are only representative of the latest non-OK events, which I tend to think of as "incidents". 

## Possible Solution

Allow the API to treat `/incidents` as an alias for `/events` endpoint.

## Context

Prior to the introduction of metric check type, events were only generated when a check result indicated a non-OK state. This is why the `/events` endpoint only exposes a subset of all events.

As a result, explaining this behavior of the `/events` endpoint involves a lot of hand-waving. I believe that exposing the same information via `/incidents` endpoint would make this much easier to explain.

cwjohnston:
## Description

Allow access to /events API endpoints via /incidents route.

## Related Issue

Closes #1690 

## Motivation and Context

Provides a better named API route for access to non-OK event data

## How Has This Been Tested?

Added unit tests

## Types of changes

## Checklist:

### DIFF
--- a/lib/sensu/api/routes/events.rb
+++ b/lib/sensu/api/routes/events.rb
@@ -6,9 +6,9 @@ module Sensu
       module Events
         include Utilities::ResolveEvent
 
-        EVENTS_URI = /^\/events$/
-        EVENTS_CLIENT_URI = /^\/events\/([\w\.-]+)$/
-        EVENT_URI = /^\/events\/([\w\.-]+)\/([\w\.-]+)$/
+        EVENTS_URI = /^\/(?:events|incidents)$/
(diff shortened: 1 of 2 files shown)


portertech:
90cc80c0-7afb-11e4-9130-b5de5f5b2ebf

PR #1703, opened 2017-07-11
[WIP] Event check previous output and status for more context

portertech:


portertech:
Still need to add tests.

majormoses:
Seems more or less how it should work, so tests and documentation about this need to be added.

majormoses:
the test results are surprising...

portertech:
Going to shelve this one for now.

PR #1712, opened 2017-07-27
Acknowledge keepalives on next EM tick

portertech:
We discovered a situation where poor Redis performance can negatively impact keepalive processing, potentially triggering a compounding failure that Sensu is unable to recover from (unless Redis is flushed and its performance improves). Keepalive acknowledgments were nested within several Redis commands/operations, in the event of degrading Redis performance, the keepalive queue would back up, triggering keepalive events that produced additional Redis commands/operations, further taxing Redis, exacerbating the issue. Moving keepalive acknowledgments out of the nested Redis commands/operations to the next tick of the EventMachine reactor removes any guarantee that the keepalive successfully updated the client registry, however, it avoids this failure situation entirely.

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -187,14 +187,14 @@ module Sensu
           @logger.debug("received keepalive", :message => message)
           begin
             client = Sensu::JSON.load(message)
-            update_client_registry(client) do
-              @transport.ack(message_info)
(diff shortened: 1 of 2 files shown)


cwjohnston:
I believe this should close https://github.com/sensu/sensu-transport/issues/38

Issue #1724, opened 2017-08-18
Optionally extend Client socket HTTP Basic Auth to info and results endpoints


stuwil:
## Expected Behavior
Feature request: the ability to restrict the other two current sensu-client HTTP endpoints (`/info` and `/results`) using the basic auth credentials specified in the `{ "client": { "http_socket": {} } }` configuration scope. If enabled, requests to those endpoints should require the same credentials used to protect the `/settings` endpoint.

## Current Behavior
The username and password specified in the `http_socket` settings only apply to requests to the `/settings` endpoint.

## Possible Solution
An additional `protect_all_endpoints` boolean can be added to the `{ "client": { "http_socket": {} } }` configuration scope, with a default of false. When enabled, the existing `authorization?` method can be used to determine if the requests to `/info` and `/results` should be processed.

## Context
Basic Auth would be useful with the `/results` endpoint in particular, in environments where selective processes should have the ability to relay check results.

PR #1726, opened 2017-08-18
Extend basic auth to info and results endpoints

stuwil:
## Description
Adds a `protect_all_endpoints` boolean flag to the `http_socket` settings as described in #1724, allowing for the other resource endpoints (`info` and `results`) to be protected by basic authorization along with `settings`.

Includes a one-liner fix for the response sent when a method is not allowed. Let me know if it's preferable to put this in a separate PR.

## Related Issue
#1724

## Motivation and Context
Extends the existing basic auth protection in a backwards-compatible manner to the other endpoints. It's useful with the /results endpoint in particular, in environments where selective processes should have the ability to relay check results.

## How Has This Been Tested?
Manually tested making requests without the option set, and with the boolean set to both true and false. Could use some pointers on writing the test spec - I didn't see an obvious way of overwriting the client settings from within the spec.

## Types of changes

## Checklist:
- [ ] All new and existing tests passed.

### DIFF
--- a/lib/sensu/client/http_socket.rb
+++ b/lib/sensu/client/http_socket.rb
@@ -188,7 +188,7 @@ module Sensu
               :http_request_uri => @http_request_uri
             })
             send_response(405, "Method Not Allowed", {
-              :response => "Valid methods for this endpoint: #{reqdef['methods'].keys}"
+              :response => "Valid methods for this endpoint: #{endpoint['methods'].keys}"
             })
           end
         else
--- a/lib/sensu/client/http_socket.rb
+++ b/lib/sensu/client/http_socket.rb
@@ -87,6 +87,14 @@ module Sensu
         false
       end
 
+      def unauthorized_response
+        @logger.warn("http socket refusing to serve unauthorized request")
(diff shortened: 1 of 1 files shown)


portertech:
@stuwil this looks great! We do however need some tests, any chance you can find some time to write two quick tests up for authentication on the newly protected endpoints? Should be able to copy https://github.com/sensu/sensu/blob/7c49fee736fa4866043e7a37d77c09f7149fc9f0/spec/client/http_socket_spec.rb#L53-L83 and make a few modifications.

lib/sensu/client/http_socket.rb
```diff
@@ -100,6 +108,9 @@ def send_response(status, status_string, content)
       end
 
       def process_request_info
+        if @settings[:client][:http_socket][:protect_all_endpoints]
```
portertech: I'll go ahead and create a sensu-settings PR to add validation for `protect_all_endpoints`.

portertech:
Going to go ahead an merge this into release/1.1, as I am going to follow up with a spec and a sensu-settings bump (adding validation) 👍 

@stuwil thank you!!!

jackyeahs

Issue #1680, opened 2017-06-06
Feature Request: Add the ability to add context to an event on non OK statuses


majormoses:
## Expected Behavior
I have often felt sensu makes it hard to provide meaningful context on alerts. I have thought about this and think that we need to support adding additional context outside of check logic. To better explain I am going to give an example and then outline some use cases.

When monitoring something like high CPU % it would be helpful to before sending the alert to say include the top 10 processes by CPU. In this case its not as simple as just include a link to a graph as most people graph overall CPU usage and maybe certain processes but not for all processes on the system. 

Another Example (added based on discussions below) would be healthchecking a service over http, if the service is returning a non ok then we might want to add context such as heap usage if its a java application.

Here is my proposal as to the types of use cases and how a user might set this up.

in check:
```json
{
  "context": {
    "type": "command",
    "retrieve": "some_command_here"
  }
}
```

I think we should support at least 3 types:
- command
- link
- arbitrary ruby

## Current Behavior
In order to do this every check must be responsible for providing its own relevant context (which is pretty insane when you consider all the use cases for say check-http....

## Possible Solution
Enhance sensu to allow adding this extra context.

## Context
How appropriate you ask for this...

mcbsdEU:
In our environment, we prefer to use our scripts in the check definitions. Community plugins are nice, but have lack of flexibility when adjustment is needed as every env. is often noticeably unique.
I am just thinking, that I would enhance the script so after detecting W|C threshold it would print top 10 processes in addition to common output like "CPU has reached....". This combined std out is automatically visible in ticketing system to which we sent events.
From my perspective, meaningful context is as much meaningful as you can make it.
But I am not saying that there are not use cases where this feature would have visible benefits...just can't think of them right now.
My 5 cents...cheers.

majormoses:
Agreed that when there is clear context it makes sense to put it in the check directly but there are cases where the context is entirely independent of check logic. For example check-http.rb when its failing it might be helpful to include say heap information if its a java app. Rather than creating a bunch of different almost identical scripts just to add different context seems less desirable than adding the ability to allow this to be defined in the check definition.

mcbsdEU:
I see...
Just thinking loud...I would extend the script and add options like "--heap"  when heap info is needed. You end up with same script, just different calls. We have modifications like that if we want to force the script to use ipv6 instead of v4 etc.
Depends on complexity...

majormoses:
[...]

PR #1748, opened 2017-09-22
Check Hooks

portertech:
The Sensu client’s primary function is to execute check commands and publish their results for processing, allowing a Sensu server to take the appropriate actions. The practice of using Sensu for automated data gathering for incident triage continues to gain traction amongst Sensu users. Although this practice is achievable, Sensu does not currently make it easy, as all actions must be orchestrated by a Sensu server.

This pull request introduce check hooks, commands run by the Sensu client in response to the result of the check command execution. The Sensu client will execute the appropriate configured hook command, depending on the check execution status (e.g. 1). Valid hook names include (in order of precedence): "1"-"255", "ok", "warning", "critical", "unknown", and "non-zero". The check hook command output, status, executed timestamp, and duration are captured and published in the check result.

Example check hook configuration:

``` json
{
  "checks": {
    "my_check": {
      "command": "echo WARNING && exit 1",
      "...": "...",
      "hooks": {
        "ok": "echo ALL IS WELL",
        "warning": "echo DANGER WILL ROBINSON!",
        "critical": "ps aux"
      }
    }
  }
}
```

Example check result hook data:

```
{
  "hooks": {
    "warning": {
      "command": "echo HOOKED",
      "timeout": 5,
      "output": "HOOKED\n",
      "status": 0,
      "executed": 1506049908,
      "duration": 0.004
    }
  }
}
```

Closes https://github.com/sensu/sensu/issues/1680

### DIFF
--- a/lib/sensu/client/process.rb
+++ b/lib/sensu/client/process.rb
@@ -117,7 +117,13 @@ module Sensu
       # substituted with the associated client attribute values, via
       # `substitute_tokens()`. If there are unmatched check attribute
       # value tokens, the check hook will not be executed, instead the
-      # hook command output will be set to report the unmatched tokens.
+      # hook command output will be set to report the unmatched
+      # tokens. Hook commands may expect/read and utilize JSON
+      # serialized Sensu client and check data via STDIN, if the hook
+      # definition includes `"stdin": true` (default is `false`). A
+      # hook may have a configured execution timeout, e.g. `"timeout":
+      # 30`, if one is not specified, the timeout defaults to 60
+      # seconds.
       #
       # @param check [Hash]
       # @yield [check] callback/block called after executing the check
@@ -134,7 +140,7 @@ module Sensu
           started = Time.now.to_f
           hook[:executed] = started.to_i
           if unmatched_tokens.empty?
-            options = {:timeout => hook[:timeout]}
+            options = {:timeout => hook.fetch(:timeout, 60)}
             if hook[:stdin]
               options[:data] = Sensu::JSON.dump({
                 :client => @settings[:client],
--- a/lib/sensu/client/process.rb
+++ b/lib/sensu/client/process.rb
@@ -113,7 +113,7 @@ module Sensu
       def execute_check_hook(check)
         @logger.debug("attempting to execute check hook", :check => check)
         severity = SEVERITIES[check[:status]] || "unknown"
-        hook = check[:hooks][check[:status]] || check[:hooks][severity]
+        hook = check[:hooks][check[:status]] || check[:hooks][severity.to_sym]
         if hook.nil? && check[:status] != 0
           hook = check[:hooks]["non-zero"]
         end
--- a/lib/sensu/client/process.rb
+++ b/lib/sensu/client/process.rb
@@ -110,6 +110,16 @@ module Sensu
         [check[:source], check[:name]].compact.join(":")
       end
 
+      # Execute a check hook, capturing its output (STDOUT/ERR),
+      # exit status code, executed timestamp, and duration. This
+      # method determines which hook command to run by inspecting the
+      # check execution status. Check hook command tokens are
+      # substituted with the associated client attribute values, via
+      # `substitute_tokens()`. If there are unmatched check attribute
+      # value tokens, the check hook will not be executed, instead the
+      # hook command output will be set to report the unmatched tokens.
+      #
+      # @param check [Hash]
       def execute_check_hook(check)
         @logger.debug("attempting to execute check hook", :check => check)
         severity = SEVERITIES[check[:status]] || "unknown"
--- a/lib/sensu/client/process.rb
+++ b/lib/sensu/client/process.rb
@@ -130,11 +130,18 @@ module Sensu
           hook = check[:hooks]["non-zero"]
         end
         if hook
-          command, unmatched_tokens = substitute_tokens(hook[:command], @settings[:client])
+          command, unmatched_tokens = substitute_tokens(hook[:command].dup, @settings[:client])
           started = Time.now.to_f
           hook[:executed] = started.to_i
           if unmatched_tokens.empty?
-            Spawn.process(command, :timeout => hook[:timeout]) do |output, status|
+            options = {:timeout => hook[:timeout]}
+            if hook[:stdin]
+              options[:data] = Sensu::JSON.dump({
+                :client => @settings[:client],
+                :check => check
+              })
+            end
+            Spawn.process(command, options) do |output, status|
               hook[:duration] = ("%.3f" % (Time.now.to_f - started)).to_f
               hook[:output] = output
               hook[:status] = status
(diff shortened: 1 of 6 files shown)


portertech:
Now to get feedback from users 🎉 

YARDOC and sensu-settings validation to come tomorrow.

portertech:
For posterity, I did consider using an array of hook commands, e.g. `"warning": [{"command": "..."}, {"command": "..."}]`, in order to support multiple sequential executions (if the previous exited 0), however, it adds unnecessary complexity as the same can be achieved with existing (and more powerful) shell operators, e.g. `&&` and `||`.

majormoses:
@portertech agreed that really we can just rely on shell/ruby operators to accomplish different that. 

Regarding different hook types when we discussed this (yes a while ago) I could think of two basic types of hooks: command and http(s). Are you thinking that we would just write a ruby, shell, python, etc script to retrieve data (such as $HOST:PORT/metrics) or should we have a bit more native approach?

If so I was thinking something along these lines:
```json
{
  "checks": {
    "my_check": {
     "hook_type" : "(command|http)"
      "hook_exec": "echo WARNING && exit 1",
      "...": "...",
      "hooks": {
        "ok": "echo ALL IS WELL",
        "warning": "echo DANGER WILL ROBINSON!",
        "critical": "ps aux"
      }
    }
  }
}
```

Basically just whether I need to either write wrapper scripts, use `curl $HOST:$PORT/$PATH`, or if we should support an HTTP endpoint natively.

majormoses:
To be clear I am not saying we should support it natively just asking the question. An argument to be made against it is that now we basically have to support every header option under the sun as [someone will ask for it](https://github.com/hashicorp/consul/issues/1184).

portertech:
@majormoses I am indeed thinking that we will write and use shell, ruby, golang, python, etc scripts to retrieve data and do things 👍

majormoses:
> "echo DANGER WILL ROBINSON!"

I am not gonna lie I kinda want to add this to all my monitoring now.

spec/config.json
```diff
@@ -196,6 +196,19 @@
...
+      "command": "echo FAIL && exit 1",
+      "subscribers": [
+        "test"
+      ],
+      "hooks": {
+          "warning": {
+              "command": "echo HOOKED",
+              "timeout": 5
+          }
+      },
```
davideauee: Note that the commit message is describing a different config format.

Also, a default timeout should probably be enforced or the attribute made mandatory.
If "Spawn.process" does already enforce something then it should probably be mentioned in the comments/docs

lib/sensu/client/process.rb
```diff
@@ -110,6 +110,46 @@ def check_in_progress_key(check)
...
+      #   hook (if any).
+      def execute_check_hook(check)
+        @logger.debug("attempting to execute check hook", :check => check)
+        severity = SEVERITIES[check[:status]] || "unknown"
+        hook = check[:hooks][check[:status].to_s] || check[:hooks][severity.to_sym]
+        if hook.nil? && check[:status] != 0
+          hook = check[:hooks]["non-zero"]
+        end
+        if hook
+          command, unmatched_tokens = substitute_tokens(hook[:command], @settings[:client])
```
davideauee: in addition to leveraging client config tokens, it could be convenient to provide/document a mechanism to pass all or parts of check result event to the hook's command as well.

calebhailey: Oh, that's a good idea.

davideauee: Glad to hear 😄,

...  because another thought that I cannot quite shake is that this is new feature is bound to be used to cannabelize server side handlers/extensions. For example, an easy temptation would be to use those hooks to fanout processing logic that can get problematic when funneled through a reduced number of sensu server instances.

I fully understand that what follows is quite a stretch from the intent of this pull request but... 

In effect what we have here is a poor's man client side handler but without the supporting framework that server side handlers/extensions enjoy (configuration, logging, defined interface, etc.).

And if we follow this train of thought to it logical conclusion... should this hook concept really be about allowing the client to run handlers/extension to process check result payloads?

It would arguably be a heavier process than the proposed desgin (at the very least you would likely be bound to use ruby) but it would also be more powerfully integrated, more inlined with the existing patterns, would not require to specify a separate command/handler for each status value, etc...

I feel a bit like the devil's advocate here but still,... food for thoughts.

d.

portertech: @davideauee thank you for the feedback!

I could add `"stdin": true` as a hook option (default to `false`), and have the Sensu client write the check definition with the populated `output`, `status`, etc to the hook process STDIN 👍 

Providing "semi event data" to the hook process STDIN is a possibility, however, it would be lacking the context constructed by the Sensu server, e.g. history, flapping, etc, and functionality like silencing would not work. I personally do not like the idea of overloading Check Hooks.

spec/config.json
portertech: We can introduce a default timeout for hooks, as it will not break backwards compatibility 👍

lib/sensu/client/process.rb
davideauee: leveraging STDIN would do the trick indeed 👍 .

I see what you mean by not wanting to "overload" Check Hooks as the intent here is to provide something both light weight and pragmatic, but it arguably comes at a cost of some amount of shoehorning.   

looking at the pros of each approach:
- client-side handlers:
   - more consistent with already established and well understood patterns 
   - well defined interfaces to feed data (configs, event data, etc.) so no need to shoehorn it through more esoteric means (config tokens, split triggers based on status, stdin, etc.)
   - could have its own config
   - better at promoting reuse
- check hooks:
   - not limited to ruby
   - easier to leverage existing local admin/ops tools (no need for a ruby shim) 
   - should be faster (at the very least, no need to bring up a ruby vm)

spec/config.json
portertech: @davideauee update the PR to set/use a default hook timeout of 60 seconds.

portertech:
@davideauee I have update the PR with hook "stdin" support, an option to write JSON serialized Sensu client and check data to the hook command process via STDIN. The JSON data closely resembles the Sensu event data that goes through the event pipeline on Sensu servers. I avoided any reference to "events", as check hooks are not intended to be client side event handlers.

portertech:
We have an opportunity to add "stdin" to checks, providing optional client and check data to check plugins.

portertech:
50-caliber-rambo-kitty-assault-unicorn

MrMoronIV:
So this is implemented now and this is the way to define them?


{
  "checks": {
    "my_check": {
      "command": "echo WARNING && exit 1",
      "...": "...",
      "hooks": {
        "ok": {"command": "echo ALL IS WELL"},
        "warning": {"command": "echo DANGER WILL ROBINSON!"},
        "critical": {"command": "ps aux"}
      }
    }
  }
}

majormoses:
Its not in the example docs (would love a PR to update them:
```
{
  "checks": {
    "my_check": {
      "command": "echo WARNING && exit 1",
      "...": "...",
      "hooks": {
        "ok": {
          "command": "echo ALL IS WELL",
          "timeout": 1
        },
        "warning": {
          "command": "echo DANGER WILL ROBINSON!",
          "timeout": 5
          },
        "critical": {
          "command": "ps aux",
          "timeout": 10
          }
      }
    }
  }
}```

PR #1754, opened 2017-09-26
1.1.0

portertech:
### Features

Check hooks, commands run by the Sensu client in response to the result of                                                                                                                                                                                     the check command execution. The Sensu client will execute the appropriate
configured hook command, depending on the check execution status (e.g. 1).                                                                                                                                                                                     Valid hook names include (in order of precedence): "1"-"255", "ok",
"warning", "critical", "unknown", and "non-zero". The check hook command                                                                                                                                                                                       output, status, executed timestamp, and duration are captured and                                                                                                                                                                                              published in the check result. Check hook commands can optionally receive
JSON serialized Sensu client and check definition data via STDIN.

Check STDIN. A boolean check definition attribute, `"stdin"`, when set to                                                                                                                                                                                      `true` instructs the Sensu client to write JSON serialized Sensu client
and check definition data to the check command process STDIN. This
attribute cannot be used with existing Sensu check plugins, nor Nagios                                                                                                                                                                                         plugins etc, as the Sensu client will wait indefinitely for the check
process to read and close STDIN.

Splayed proxy check request publishing. Users can now splay proxy check
requests (optional), evenly, over a window of time, determined by the                                                                                                                                                                                          check interval and a configurable splay coverage percentage. For
example, if a check has an interval of 60s and a configured splay                                                                                                                                                                                              coverage of 90%, its proxy check requests would be splayed evenly over a
time window of 60s * 90%, 54s, leaving 6s for the last proxy check
execution before the the next round of proxy check requests for the same                                                                                                                                                                                       check. Proxy check request splayed publishing can be configured with two
new check definition attributes, within the `proxy_requests` scope,                                                                                                                                                                                            `splay` (boolean) to enable it and `splay_coverage` (integer percentage,                                                                                                                                                                                       defaults to `90`).

Configurable check output truncation (for storage in Redis). Check output
truncation can be manually enabled/disabled with the check definition                                                                                                                                                                                          attribute "truncate_output", e.g.`"truncate_output": false`. The output
truncation length can be configured with the check definition attribute
"truncate_output_length", e.g. `"truncate_output_length": 1024`. Check                                                                                                                                                                                         output truncation is still enabled by default for metric checks, with
`"type": "metric"`.

Sensu client HTTP socket basic authentication can how be applied to all
endpoints (not just `/settings`), via the client definition http_socket                                                                                                                                                                                        attribute "protect_all_endpoints", e.g. `"protect_all_endpoints": true`.

### Other

Improved check TTL monitoring performance.

The Sensu extension run log event log level is now set to debug (instead                                                                                                                                                                                       of info) when the run output is empty and the status is 0.

### Fixes

Added initial timestamp to proxy client definitions. The Uchiwa and Sensu                                                                                                                                                                                      dashboards will no longer display "Invalid Date".

Deleting check history when deleting an associated check result.


portertech:
dog-dog

PR #1769, opened 2017-10-18
API log request ID and response body content (debug only)

portertech:
This pull-request adds API request ID (header) and response body content logging.

The Sensu API will generate a request ID if the HTTP client does not provide one. The request ID can be used to trace a HTTP request via the Sensu API log.

We are back-porting these changes to get them into an Enterprise patch release sooner.


portertech:
Ignoring the Ruby 2.2 Travis CI failure, it's due to a Bundler version issue that was fixed in a later release.

portertech:
caremad

Issue #1773, opened 2017-10-25
Check hooks not firing


absolutejam:
After upgrading to 1.1.1-1 (server and API) for check hooks, I'm struggling to get them to work.

## Expected Behavior
Check hook should fire and display information in the `/checks` API and Uchiwa.

Example check:
```json
{
  "checks": {
    "disk-free": {
      "command": "C:\\\\opt\\\\sensu\\\\embedded\\\\bin\\\\ruby C:\\\\opt\\\\sensu\\\\check-scripts\\\\check-windows-disk.rb -w :::vars.disk.warn|80::: -c :::vars.disk.crit|90:::",
      "description": "Checks how full disks are",
      "handlers": [
        "default"
      ],
      "hooks": {
        "2": {
          "command": "echo test",
          "timeout": 30
        }
      },
      "interval": 60,
      "occurrences": 1,
      "refresh": 3600,
      "subscribers": [
        "winservers"
      ]
    }
  }
}
```
(I have tried this with `"non-zero"` but also tried `2` to see if this worked as an explicit value).

When the check fires (or I fire it via. API), there is no mention of the check hook result, only the check hook info from the check definitioon

## Current Behavior
No hook result is recorded. This is the same on Linux and Windows hosts.

## Your Environment
* Sensu version used: 1.1.1-1
* Operating System and version (e.g. Ubuntu 14.04): Ubuntu 16.04, Windows Server 2012 R2, Windows Server 2016
* Transport and runtime versions (e.g. RabbitMQ 3.6.2 on Erlang 18.2): Redis currently (Redis server v=3.2.11 sha=00000000:0 malloc=jemalloc-4.0.3 bits=64 build=60be07af25e933f5)

portertech:
Expected behaviour is:

Check hook should fire and display information in the `/results/:client/:check` and `/events/:client/:check` APIs and Uchiwa.

We discussed this a little on the Sensu Community Slack, expecting an update soon.

epierotto:
Same issue here.
I'm running sensu 1.1.1 on CentOS
This is the hook definition inside the check:
```
"hooks": {
    "non-zero": {
    "command": "df -h"
  }
},
```
I check `/results/:client/:check` on API and the are no signs of hook execution.
Also changed `LOG_LEVEL=debug` on the client and I can see pretty much the same json that the API has.

portertech:
[...]

PR #1784, opened 2017-11-24
Fix Check Hooks

portertech:
Sensu check request JSON parsing produces a hash with symbolized keys, therefor symbolized keys must be used for hooks. This pull-request fixes hooks named after status codes, e.g. `1`, and `non-zero`.

Previously, if a Sensu client had a local check definition, the settings loader provided both symbol and string keys and hooks worked. Hooks with severity names, e.g. `warning`, also worked, because Sensu was converting them to symbols.

Closes https://github.com/sensu/sensu/issues/1773


zbintliff:
We can't find this version in the yum repo. Is it available?

PR #1790, opened 2017-12-01
Scheduled maintenance event silencing

portertech:
Scheduled maintenance. Sensu currently gives users the ability to silence a check and/or client subscriptions, stopping related Sensu events from being handled (e.g. no notifications). This pull request adds the ability to silence a check and/or client subscriptions at a predetermined time (`begin` epoch timestamp), enabling users to silence events for scheduled maintenance windows in advance.

### DIFF
--- a/lib/sensu/api/routes/silenced.rb
+++ b/lib/sensu/api/routes/silenced.rb
@@ -59,6 +59,7 @@ module Sensu
               subscription = data.fetch(:subscription, "*")
               check = data.fetch(:check, "*")
               silenced_id = "#{subscription}:#{check}"
+              timestamp = Time.now.to_i
               silenced_info = {
                 :id => silenced_id,
                 :subscription => data[:subscription],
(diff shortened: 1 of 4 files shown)


portertech:
cheeky_excited

portertech:
We could also add `end` (epoch timestamp) and calculate the correct expire time in seconds, however, this could create situations where there is a conflict between `end` and `expire`, which one wins etc?

portertech:
@cwjohnston @amdprophet @palourde I would love to get your input on this feature.

cwjohnston:
@portertech I like the idea of scheduling silencing entries this way, but I think there's something missing. Can we annotate the event data to indicate that the silencing was scheduled in advance? I think this detail is needed for folks to generate SLA compliance reports, i.e. was an event silenced because we had planned ahead (scheduled) or because we got paged unexpectedly (unscheduled)?

portertech:
@cwjohnston hmm, good idea. I'm concerned it would require making significant changes to the silencing event data, format/structure. There could be multiple entries for an event, some scheduled, so not. We could just go with a single event attribute, e.g. "scheduled_maintenance" (but this is yet another term). We would set the attribute at:

https://github.com/sensu/sensu/blob/81b0e74b24f97a82cbf4c29132619b0e07f16abb/lib/sensu/server/process.rb#L489-L492

portertech:
@cwjohnston this would also require a feature rework, couldn't default begin to the current timestamp.

portertech:
@cwjohnston I am going to update the implementation to not set `begin` to now when not specified. This will support your desired functionality in the future, but not implement it in this pr.

Issue #1799, opened 2017-12-18
FR: Allow the Client TCP/UDP JSON Socket to be disabled


jaredledvina:
I would like to be able to completely disable the TCP/UDP client socket and only expose the HTTP socket.

## Possible Solution
I think the easiest way to do this is to add an `enabled` option to this section: https://github.com/sensu/sensu/blob/master/lib/sensu/client/process.rb#L464-L480 similar to how the `setup_http_socket` function uses that option. 

## Context
In my environment, we are looking to disable the TCP/UDP socket as it can not require authentication whereas the HTTP socket can. 

## Your Environment
* Sensu version used: 1.2.0
* Active Sensu extensions: None
* Operating System and version: CentOS 6 and CentOS7
* Transport and runtime versions: RabbitMQ 3.6.5 and Erlang 19.1.5

cwjohnston:
Closed by #1800 👍

PR #1800, opened 2017-12-18
TCP/UDP socket supports enabled option

jaredledvina:
This PR adds the ability for the client's TCP/UDP socket to be disabled. 

## Description
Mimicking how the HTTP socket is setup, this change will allow users to disable the TCP and UDP socket completely by setting `enabled: false` in their socket configuration. 

## Related Issue
https://github.com/sensu/sensu/issues/1799

## Motivation and Context
In my environment, the TCP/UDP socket is deemed a security risk as it doesn't not support authentication (whereas the HTTP socket does). This change will allow me to completely disable it and rely on the HTTP socket instead. 

## How Has This Been Tested?
I could actually use some help with testing this out. Are there any testing docs I could run through to verify this doesn't cause a regression? 

## Types of changes
- [X] New feature (non-breaking change which adds functionality)

## Checklist:
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.
- [X] All new and existing tests passed.

### DIFF
--- a/lib/sensu/client/process.rb
+++ b/lib/sensu/client/process.rb
@@ -455,27 +455,33 @@ module Sensu
       # Setup the Sensu client JSON socket, for external check result
       # input. By default, the client socket is bound to localhost on
       # TCP & UDP port 3030. The socket can be configured via the
-      # client definition, `:socket` with `:bind` and `:port`. The
-      # current instance of the Sensu logger, settings, and transport
+      # client definition, `:socket` with `:bind` and `:port`. Users can opt-out
+      # of using the TCP and UDP socket by setting `:enabled` to `false`.
+      # The current instance of the Sensu logger, settings, and transport
       # are passed to the socket handler, `Sensu::Client::Socket`. The
       # TCP socket server signature (Fixnum) and UDP connection object
       # are stored in `@sockets`, so that they can be managed
       # elsewhere, eg. `close_sockets()`.
       def setup_json_socket
         options = @settings[:client][:socket] || Hash.new
+        options[:enabled] ||= true
         options[:bind] ||= "127.0.0.1"
         options[:port] ||= 3030
-        @logger.debug("binding client tcp and udp sockets", :options => options)
-        @sockets << EM::start_server(options[:bind], options[:port], Socket) do |socket|
-          socket.logger = @logger
-          socket.settings = @settings
-          socket.transport = @transport
-        end
-        @sockets << EM::open_datagram_socket(options[:bind], options[:port], Socket) do |socket|
-          socket.logger = @logger
-          socket.settings = @settings
-          socket.transport = @transport
-          socket.protocol = :udp
+        if options[:enabled]
+          @logger.debug("binding client tcp and udp sockets", :options => options)
+          @sockets << EM::start_server(options[:bind], options[:port], Socket) do |socket|
+            socket.logger = @logger
+            socket.settings = @settings
+            socket.transport = @transport
+          end
+          @sockets << EM::open_datagram_socket(options[:bind], options[:port], Socket) do |socket|
+            socket.logger = @logger
+            socket.settings = @settings
+            socket.transport = @transport
+            socket.protocol = :udp
+          end
+        else
+            @logger.info("client tcp/udp socket disabled per configuration")
         end
       end
 


lib/sensu/client/process.rb
```diff
@@ -465,17 +466,19 @@ def setup_json_socket
...
-          socket.logger = @logger
-          socket.settings = @settings
-          socket.transport = @transport
-        end
-        @sockets << EM::open_datagram_socket(options[:bind], options[:port], Socket) do |socket|
-          socket.logger = @logger
-          socket.settings = @settings
-          socket.transport = @transport
-          socket.protocol = :udp
+        unless options[:enabled] == false
```
majormoses: few things, rather than checking a negative value and then negating it, I generally find it easier to read as checking for the positive value instead. Also I think that disabling this should be logged so I'd write it something like this:
```ruby
if options[:enabled]
  # insert existing logic
else
  # log that client socket is being disabled per configuration.
end
```

majormoses: Another option would be to do something like this:
```ruby
if !options[:enabled]
  # log socket disabled
else
  # insert existing logic
end
```

majormoses: The bottom one forces negation to ensure that `nil` values are evaluated as true (current default)

majormoses: The following illustrates the difference between the two:
```ruby
$ irb
irb(main):001:0> options = {}
=> {}
irb(main):002:0> if options[:foo] then p 'true' else p 'false' end
"false"
=> "false"
irb(main):003:0> if !options[:foo] then p 'true' else p 'false' end
"true"
=> "true"
irb(main):004:0> if options[:foo] || options[:foo].nil? then p 'true' else p 'false' end
"true"
=> "true"
```

You could also include an `|| options[:enabled].nil?` to catch this

portertech: `if options[:foo] || options[:foo].nil?` is the most verbose and imo happens to be much easier to ready and understand.

jaredledvina: Hey @majormoses  and @portertech, thanks for the quick review! I pushed a change based on your feedback. My Ruby skills are very basic so, let me know if anything looks even slightly wrong. 

Should I also update `setup_http_socket` to use a similar logic/logging setup?

lib/sensu/client/process.rb
```diff
@@ -455,27 +455,33 @@ def setup_standalone
...
-          socket.logger = @logger
-          socket.settings = @settings
-          socket.transport = @transport
-        end
-        @sockets << EM::open_datagram_socket(options[:bind], options[:port], Socket) do |socket|
-          socket.logger = @logger
-          socket.settings = @settings
-          socket.transport = @transport
-          socket.protocol = :udp
+        if options[:enabled] || options[:enabled].nil
```
majormoses: This should be `.nil?` read below for more flavor

In ruby there is a convention when asking questions such as if something is `nil` to use a question mark in the function name. This provides more clarity on its use. Take for example:
```ruby
$ irb
irb(main):001:0> a = nil
=> nil
irb(main):003:0> b = a.nil?
=> true
irb(main):004:0> b = nil?
=> false
```

majormoses: Also since you defined a default of true above the check for `nil` becomes less important as the only way this logic would be needed if someone explicitly set their config to `nil`. I don't know why someone would do that as if they wanted to explicitly set then would set `(true|false)`.

lib/sensu/client/process.rb
majormoses: @portertech your call but I think it's worth the minimal effort to ensure that they act the same way. We could also split them up as 2 separate PRs if you'd feel better about that.

lib/sensu/client/process.rb
jaredledvina: Ah cool, yeah the `?` syntax makes sense. Does it make sense for me to define the default or should I scrap that and check for it being `nil`? Personally, I'm thinking being completely explicit with the defaults makes sense but, I'm game for whatever is more standard here.

lib/sensu/client/process.rb
```diff
@@ -455,27 +455,33 @@ def setup_standalone
...
+            socket.transport = @transport
+          end
+          @sockets << EM::open_datagram_socket(options[:bind], options[:port], Socket) do |socket|
+            socket.logger = @logger
+            socket.settings = @settings
+            socket.transport = @transport
+            socket.protocol = :udp
+          end
+        else
+            @logger.info("client tcp/udp socket disabled per configuration")
```
jaredledvina: Is info level logging fine or does this feel more like a `debug` message?

lib/sensu/client/process.rb
```diff
@@ -455,27 +455,33 @@ def setup_standalone
...
+      # client definition, `:socket` with `:bind` and `:port`. Users can opt-out
+      # of using the TCP and UDP socket by setting `:enabled` to `false`.
+      # The current instance of the Sensu logger, settings, and transport
       # are passed to the socket handler, `Sensu::Client::Socket`. The
       # TCP socket server signature (Fixnum) and UDP connection object
       # are stored in `@sockets`, so that they can be managed
       # elsewhere, eg. `close_sockets()`.
       def setup_json_socket
         options = @settings[:client][:socket] || Hash.new
+        options[:enabled] ||= true
```
jaredledvina: Went with being explicit with the default but, if we'd rather I don't define this and check for the `nil?` instead, happy to switch.

lib/sensu/client/process.rb
majormoses: I think info is appropriate, it's not default and I find most places with high volume end up using warn or higher by default and then drop to info or debug for troubleshooting. I can't speak beyond my experiences though. I will support whatever sean decides.

lib/sensu/client/process.rb
majormoses: well following the existing pattern they do not check for `nil` and relies on defaults explicitly defined. I am :+1: on following the existing so I'd keep this and probably remove the conditional for `nil?`.

lib/sensu/client/process.rb
majormoses: commented elsewhere but I'd prefer explicit default and drop the conditional.

cwjohnston:
Closes #1799

portertech:
This is rad 👍 Thank you!



This feature requires configuration validation for "enabled", I'll take ownership of a sensu-settings PR for this!

majormoses:
After this releases can we add the docs that were mistakenly merged here early? https://github.com/sensu/sensu-docs/pull/678

jaredledvina:
Woohoo! \o/ Let me know if I can be of any additional help!

jaredledvina:
Possibly not required but, I installed https://repositories.sensuapp.org/yum-unstable/7/x86_64/sensu-1.3.0-1.el7.x86_64.rpm on 3 sensu instances today. Prior to any config changes, the `sensu-client` is listening as I expected on TCP/UDP/HTTP:
```
# ss -tulnp | grep sensu-client
udp    UNCONN     0      0      127.0.0.1:3030                  *:*                   users:(("sensu-client",pid=11610,fd=13))
tcp    LISTEN     0      100    127.0.0.1:3030                  *:*                   users:(("sensu-client",pid=11610,fd=12))
tcp    LISTEN     0      100    127.0.0.1:3031                  *:*                   users:(("sensu-client",pid=11610,fd=14))
```

And then I re-configured the `/etc/sensu/conf.d/client.json` as:
```
{
  "client": {
    "name": "blah-blah-node-blah",
    "address": "1.2.3.4",
    "subscriptions": [
      "all"
    ],
    "socket": {
      "enabled": false
    },
    "safe_mode": true,
    "keepalive": {
    }
  }
}
```

After restarting the `sensu-client`, it's now only listening w/ the HTTP socket:
```
# ss -tulnp | grep sensu-client
tcp    LISTEN     0      100    127.0.0.1:3031                  *:*                   users:(("sensu-client",pid=29927,fd=12))
```
And is very happy:
```
# curl -s  http://localhost:3031/info | jq .
{
  "sensu": {
    "version": "1.3.0"
  },
  "transport": {
    "name": "rabbitmq",
    "keepalives": {
      "messages": 0,
      "consumers": 3
    },
    "results": {
      "messages": 0,
      "consumers": 3
    },
    "connected": true
  }
}
```

portertech:
Thanks @jaredledvina!

PR #1842, opened 2018-04-11
Sensu call-home (Tessen)

portertech:
This pull-request implements a Sensu call-home mechanism. It sends anonymized data about the Sensu installation to the Tessen hosted service (Sensu Inc), on `sensu-server` startup and every 6 hours thereafter. All data reports are logged for transparency/awareness and transmitted over HTTPS. The anonymized data currently includes the flavour of Sensu (Core or Enterprise), the version, and the Sensu client and server counts.

The data will help the Sensu Team answer questions like:

- How many Sensu installations are there?

- What's the average Sensu client to server ratio?

- What Sensu versions are in use?

- How many installations use Sensu Core vs Enterprise?

jackyeahs

### Enabled by default

The Tessen client is enabled by default, users can **opt-out** via `sensu-server` configuration:

`/etc/sensu/conf.d/tessen.json`

``` json
{
  "tessen": {
    "enabled": false
  }
}
```

### Proxy Configuration

The Tessen client supports proxy configuration.

`/etc/sensu/conf.d/tessen.json`

``` json
{
  "tessen": {
    "proxy": {
      "host": "www.myproxy.com",
      "port": 4430,
      "authorization": ["username", "password"]
    }
  }
}
```

### Future

The Tessen client supports an optional "token" that can be used to sign data (opt-in). This token is not currently supported by the Tessen service, but it could be in the near future. This feature could allow Tessen to "monitor the monitor" or generate reports for Sensu users.

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -2,6 +2,7 @@ require "sensu/daemon"
 require "sensu/server/filter"
 require "sensu/server/mutate"
 require "sensu/server/handle"
+require "sensu/server/tessen"
 
 module Sensu
   module Server
@@ -1430,6 +1431,8 @@ module Sensu
               :hexdigest => @settings.hexdigest
             }
           )
+          tessen = @settings[:tessen] || {}
+          tessen_enabled = tessen.fetch(:enabled, true)
           info = {
             :id => server_id,
             :hostname => system_hostname,
@@ -1442,6 +1445,9 @@ module Sensu
               }
             },
             :sensu => sensu,
+            :tessen => {
+              :enabled => tessen_enabled
+            },
             :timestamp => Time.now.to_i
           }
           @redis.sadd("servers", server_id)
@@ -1462,6 +1468,16 @@ module Sensu
         @timers[:run] << EM::PeriodicTimer.new(10) do
           update_server_registry
         end
+      end
+
+      # Set up Tessen, the call home mechanism.
+      def setup_tessen
+        @tessen = Tessen.new(
+          :settings => @settings,
+          :logger => @logger,
+          :redis => @redis
+        )
+        @tessen.run if @tessen.enabled?
       end
 
       # Unsubscribe from transport subscriptions (all of them). This
@@ -1497,6 +1513,7 @@ module Sensu
         setup_results
         setup_task_elections
         setup_server_registry_updater
+        setup_tessen
         @state = :running
       end
 
@@ -1513,10 +1530,10 @@ module Sensu
       # Pause the Sensu server process, unless it is being paused or
       # has already been paused. The process/daemon `@state` is first
       # set to `:pausing`, to indicate that it's in progress. All run
-      # timers are cancelled, and the references are cleared. The
-      # Sensu server will unsubscribe from all transport
-      # subscriptions, relinquish any Sensu server tasks, then set the
-      # process/daemon `@state` to `:paused`.
+      # timers are cancelled, their references are cleared, and Tessen
+      # is stopped. The Sensu server will unsubscribe from all
+      # transport subscriptions, relinquish any Sensu server tasks,
+      # then set the process/daemon `@state` to `:paused`.
       def pause
         unless @state == :pausing || @state == :paused
           @state = :pausing
@@ -1524,6 +1541,7 @@ module Sensu
             timer.cancel
           end
           @timers[:run].clear
+          @tessen.stop if @tessen
           unsubscribe
           relinquish_tasks
           @state = :paused
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -1432,7 +1432,7 @@ module Sensu
             }
           )
           tessen = @settings[:tessen] || {}
-          tessen_enabled = tessen.fetch(:enabled, true)
+          tessen_enabled = tessen.fetch(:enabled, false)
           info = {
             :id => server_id,
             :hostname => system_hostname,
--- a/lib/sensu/server/tessen.rb
+++ b/lib/sensu/server/tessen.rb
@@ -70,15 +70,15 @@ module Sensu
         get_install_id do |install_id|
           get_client_count do |client_count|
             get_server_count do |server_count|
-              token = @options.fetch(:token, "")
-              type, version = get_version_info
+              identity_key = @options.fetch(:identity_key, "")
+              flavour, version = get_version_info
               timestamp = Time.now.to_i
               data = {
-                :token => token,
+                :tessen_identity_key => identity_key,
                 :install => {
                   :id => install_id,
-                  :type => type,
-                  :version => version
+                  :sensu_flavour => flavour,
+                  :sensu_version => version
                 },
                 :metrics => {
                   :points => [
--- a/lib/sensu/server/tessen.rb
+++ b/lib/sensu/server/tessen.rb
@@ -21,11 +21,17 @@ module Sensu
         @options = @settings.to_hash.fetch(:tessen, {})
       end
 
-      # Determine if Tessen is enabled (opt-out).
+      # Determine if Tessen is enabled (opt-in).
       #
       # @return [TrueClass, FalseClass]
       def enabled?
-        @options[:enabled] != false
+        enabled = @options[:enabled] == true
+        unless enabled
+          note = "tessen collects anonymized data to help inform the sensu team about installations"
+          note << " - you can opt-in via configuration: {\"tessen\": {\"enabled\": true}}"
+          @logger.info("the tessen call-home mechanism is not enabled", :note => note)
+        end
+        enabled
       end
 
       # Run Tessen, scheduling data reports (every 6h).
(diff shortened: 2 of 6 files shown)


lib/sensu/server/process.rb
```diff
@@ -1442,6 +1445,9 @@ def update_server_registry
               }
             },
             :sensu => sensu,
+            :tessen => {
```
portertech: The Sensu server registry data now includes if Tessen is enabled, this information is available via API /info 👍

lib/sensu/server/tessen.rb
```diff
@@ -0,0 +1,165 @@
...
+          timer.cancel
+        end
+        @timers.clear
+      end
+
+      # Schedule data reports, sending data to the Tessen service
+      # immediately and then every 6 hours after that.
+      def schedule_data_reports
+        note = "this data helps inform the sensu team about installations"
+        note << " - you can choose to opt-out via configuration: {\"tessen\": {\"enabled\": false}}"
```
portertech: opt-out instructions are included in a log event prior to sending the first data report.

lib/sensu/server/tessen.rb
```diff
@@ -0,0 +1,165 @@
...
+        @logger = options[:logger]
+        @redis = options[:redis]
+        @options = @settings.to_hash.fetch(:tessen, {})
+      end
+
+      # Determine if Tessen is enabled (opt-out).
+      #
+      # @return [TrueClass, FalseClass]
+      def enabled?
+        @options[:enabled] != false
```
majormoses: I feel like this is a breaking change and should be opt in for sensu 1.x and in sensu 2.x you can change the default as that is where all breaking changes should be made. I think it should be `@options[:enabled] || false` for it's initial release. We can always have the various CM tools change their defaults in a major version of the cookbook, modules, playbooks,formulas but if you want to avoid a major bump for sensu ruby and have committed to semver this is how it has to be. Also keep in mind that your data will be horribly skewed if its opt out default as it will include local development, ci, etc.

lib/sensu/server/tessen.rb
```diff
@@ -0,0 +1,165 @@
...
+      end
+
+      # Create data to be sent to the Tessen service.
+      #
+      # @return [Hash]
+      def create_data
+        get_install_id do |install_id|
+          get_client_count do |client_count|
+            get_server_count do |server_count|
+              token = @options.fetch(:token, "")
```
majormoses: can we get a more descriptive name than token? I think it's really important to be very verbose on what is being collected. The more you try to make it ambiguous the more people will feel like they don't want to share it.

lib/sensu/server/tessen.rb
```diff
@@ -0,0 +1,165 @@
...
+              type, version = get_version_info
+              timestamp = Time.now.to_i
```
majormoses: As this data will be coming from all over the globe you probably want to explicitly grab it in UTC as (sadly) some servers are not set to UTC.

lib/sensu/server/tessen.rb
```diff
@@ -0,0 +1,165 @@
...
+              data = {
+                :token => token,
+                :install => {
+                  :id => install_id,
+                  :type => type,
```
majormoses: Id like a comment on what an install type means in this context or a more verbose name. After reading through all the way to the bottom I realize this means core vs enterprise but originally I was thinking more along the lines of package management.

lib/sensu/server/tessen.rb
```diff
@@ -0,0 +1,165 @@
...
+                  :version => version
```
majormoses: I assume this in only the server version, it also might be a good idea to include a hash of number of clients by client version.

lib/sensu/server/tessen.rb
portertech: Ruby Time to_i() produces epoch

lib/sensu/server/tessen.rb
portertech: Sensu flavour?

lib/sensu/server/tessen.rb
portertech: WDYT about "tessen_token"?

majormoses: After reading the PR again I realize what the token is but we definitely want a better var name and the comments in the description about it so its intent is clear later on.

majormoses: better but I think `tessen_signing_token` makes it clearer? I'd still want a comment on the line above so it is clear as it's not gonna be something that is going to be used so explaining what it will eventually be used for is a good idea.

lib/sensu/server/tessen.rb
majormoses: I like either but I think `sensu_flavor` is clearer as `sensu_edition` might also imply a version. But either works for me.

cjchand:
Might want to look into adding auth to the proxy config. Our proxy lets traffic to certain whitelisted URIs pass without creds, but can optionally let you get to most of the Internet if you authenticate (because, auditing).

portertech:
@cjchand Ah yes, it does support authentication, I'll add it to the config example :+1:

majormoses:
I thought about this more and I think data sovereignty especially in Europe will be an issue for many. I think if you want to get more people to be OK with this you will want to offer sending to different geographic endpoints. I would rather not expose config for sending the data to an arbitrary endpoints as I worry about it being used maliciously. We could expose sending to to a specific region name and have the source code have a frozen hash of regions that map to specific endpoints to avoid misuse but provide options.

jspaleta:
Thoughts for additional data to include:
1) OS version:  
obvious security concern here

2) Can you tell if the sensu install is from a patched vendor package or official release?
maybe just a flag saying official release or patched released based on a checksum calculation. that leaks the least amount of information.

Knowing if sensu install is from sensu.io versus downstream vendor...will impact where the user expects to get their update from and if they need an update notice from sensu.io.

lib/sensu/server/tessen.rb
```diff
@@ -0,0 +1,168 @@
...
+        @logger = options[:logger]
+        @redis = options[:redis]
+        @options = @settings.to_hash.fetch(:tessen, {})
+      end
+
+      # Determine if Tessen is enabled (opt-in).
+      #
+      # @return [TrueClass, FalseClass]
+      def enabled?
+        enabled = @options[:enabled] == true
```
portertech: Tessen call-home is now opt-in for 1.x 👍

lib/sensu/server/tessen.rb
```diff
@@ -0,0 +1,168 @@
...
+      # @return [Hash]
+      def create_data
+        get_install_id do |install_id|
+          get_client_count do |client_count|
+            get_server_count do |server_count|
+              identity_key = @options.fetch(:identity_key, "")
+              flavour, version = get_version_info
+              timestamp = Time.now.to_i
+              data = {
+                :tessen_identity_key => identity_key,
```
portertech: The data keys have been updated (after feedback) to be more descriptive.

lib/sensu/server/tessen.rb
```diff
@@ -0,0 +1,168 @@
...
+                :install => {
+                  :id => install_id,
+                  :sensu_flavour => flavour,
+                  :sensu_version => version
```
majormoses: this is just the server version right? did we discuss whether we want to report on client versions as well? I am :+1: on that if its an opt-in for that specific data point.

jspaleta:
@majormoses 
no downstream is publicly packaging sensu yet afaik.. but there could be in-house version built from source (which is all redistributed packages like rpm and deb are really) not making use of the sensu.io binary packages.

I think there is value in Tessen knowing how many people are running officially distributed sensu code, and how many people are running with a downstream patch... or from source...and not just the version number. Because the sensu.io packages are more than just the sensu code, the version number is perhaps not sufficient to understand things like the security vulnerabilities of the install.  If there is a security issue in the ruby provided in the sensu.io package.. that's something sensu.io package users will need to address. but an install from source might not.

Thinking a little further... being able to tell the difference between a source install of 1.x versus a 1.x sensu.io package install might be valuable as well...as a source install doesn't include the pre-compiled ruby. The sensu.io provided rpms, for example, include a number of libraries like zlib (i think..i'd have to go check again) that duplicate functionality in the system libraries as a trade-off to providing a fully consistent official ruby stack to run on top of.

A source install (including downstream packaging as a subset of possible source installs), using a locally available ruby environment, would most likely make use of existing system libraries.. and represents a different security surface to watch out for.

majormoses:
In 1.x land now that the sensu gems are signed that should be doable as of [1.3.0](https://github.com/sensu/sensu/blob/master/CHANGELOG.md#added) not sure what the plans are from something similar for 2.x @portertech any plans there?

jspaleta:
@majormoses  I think for 2.x  at least for now... the 2.x binary packages from sensu.io provide statically linked go binaries. there should be a way to use a signed checksum to validate that what is installed is the statically linked binary expected.

or at a higher level... the sensu.io provided packages can be signed and validated. Well i know the rpms can be.. it looks like the 2.x nightly binary i have isn't signed... but it could be.  I'm sure debs have a signing and validation process as well.

majormoses:
for debian its fairly straightforward you just use `debsign` and `pgp`: https://blog.packagecloud.io/eng/2014/10/28/howto-gpg-sign-verify-deb-packages-apt-repositories/

portertech:
@majormoses @jspaleta I like these two data items, OS version and official build (boolean), as followup/roadmap issues. OS version is certainly more straightforward, at least with capturing the sensu-server OS. Determining if Sensu 1.x is an official build requires a great deal of effort for a good degree of confidence, so much of it is runtime/packaging. Sensu 2.x will be easy, as @jspaleta called out, a binary with a build signature will do it.

portertech:
I have gone ahead and created https://github.com/sensu/sensu/issues/1847 for OS family/version.

portertech:
:shipit: 

itsalive

Issue #1693, opened 2017-06-20
sensu last_ok not useful when tracking state changes


cabecada:
last_ok is really not useful when were are *only tracking state changes*, it works well otherwise.

last_ok reset to current_timestamp when check alerts changes from 0 to 2
https://github.com/sensu/sensu/blob/master/lib/sensu/server/process.rb#L576

last_ok reset to current_timestamp when check alerts changes from 2 to 0
https://github.com/sensu/sensu/blob/master/lib/sensu/server/process.rb#L583

ideally, example:

event:a
status:2
timestamp: 2017-01-01 01:01:01
last_ok: 2017-01-01 01:01:01

event:b
status:0
timestamp: 2017-01-01 02:01:01
last_ok: 2017-01-01 02:01:01

event:c
status:2
timestamp: 2017-01-01 03:01:01
last_ok: 2017-01-01 02:01:01

but we get:
event:c
status:2
timestamp: 2017-01-01 03:01:01
last_ok: 2017-01-01 03:01:01

![image](https://user-images.githubusercontent.com/20220560/27350604-9e856632-55af-11e7-925c-0e12c57965b6.png)

FYI:
we are using sensu event-bridge extension to stream events (for all statuses)

sensu: 0.26.5
ubuntu14

portertech:
For context, "last_ok reset to current_timestamp when check alerts changes from 0 to 2" is due to the fact that we only store "last_ok" when there is a current event stored in Redis, it is not tracked/stored elsewhere.

portertech:
A possible approach would be to store last_ok when storing every check result in Redis, in `store_check_result()`.

portertech:
Closed by #1856 👍

PR #1856, opened 2018-05-01
Improve event last OK

portertech:
Update check (and event) last OK when storing latest results. This PR uses a separate Redis key/value for last OK, instead of storing it within a current event (if exists).

Closes https://github.com/sensu/sensu/issues/1693


amdprophet:
Will `last_ok` still be returned with events via the API?

portertech:
@amdprophet yes, as its is still stored within current events (just no longer used as a record).

PR #1863, opened 2018-05-11
Fire registration event on startup

dsiddharth:
Current deregistration behavior makes it easy to completely delete a host and "register" it when it comes back up. This however, also deletes the host from Uchiwa and we lose all check history. 

Goal: We'd like to be able to "silence" a host when it's going down and then "unsilence" when it's back up.

Attempted to do this using a "registration" handler but that only fires when there's no client key in Redis - which isn't the case when a client was just silenced when the sensu-client process was stopping. Setting up a generic "default" handler to always unsilence a host for every event that is processed seems like it's overkill and would result in a lot of unnecessary API calls. The right solution would be to handle an event that fires when the client is starting up. This PR attempts to build out the logic to fire a "registration" event and the pass along the corresponding handler that's configured to the Sensu Server.

### DIFF
--- a/lib/sensu/client/process.rb
+++ b/lib/sensu/client/process.rb
@@ -535,6 +535,38 @@ module Sensu
         end
       end
 
+      def wait_until_running
+        retry_until_true do
+          if @state != :running
+            yield
+            true
+          end
+        end
+      end
+
+      # Create a check result intended for registration a client.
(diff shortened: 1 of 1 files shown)


mdraselahmed25:
Android

portertech:
Closing, please see https://github.com/sensu/sensu-go

PR #1882, opened 2018-07-12
Add deep_dup method, deep_dup before token substitution

### LINKED ISSUE #1855: Proxy Check with Token Substitution for the subscription retains previous values
I'm attempting to create proxy checks for different proxy clients, which are intended to be sent to different subscriptions based on the environment of the proxy client. When using token substitution in the subscribers value, the substitution seems to occur the first time sensu-server is started, but after the first token substitution happens, the subscribers value stays the same for all subsequent proxy clients. This means that the checks are being sent to the wrong subscriptions to be executed most of the time.

## Expected Behavior

When specifying a token substitution in the subscription field for a proxy check, the token substitution should happen every time the proxy check is created, not just the first time. Token Substitution for the subscription values should work the same way as it currently does for the command values.

## Steps to Reproduce

1. Create two proxy clients, each with the following definition:
```
{
  "name": "region1-router",
  "address": "192.168.1.1",
  "client_type": "router",
  "region": "region1",
  "keepalives": false,
  "subscriptions": [
    "client:region1-router"
  ]
}
```
```
{
  "name": "region2-router",
  "address": "192.168.2.1",
  "client_type": "router",
  "region": "region2",
  "keepalives": false,
  "subscriptions": [
    "client:region2-router"
  ]
}
```
2. Create a proxy check for the client
```
{
  "checks": {
    "check-router": {
      "command": "check-ping.rb -h :::address:::"
      "interval": 120,
      "subscribers": [
        "roundrobin:check-:::region:::-network"
      ],
      "enable_deprecated_filtering": false,
      "proxy_requests": {
        "client_attributes": {
          "client_type": "router"
        }
      }
    }
  }
}
```
3. Watch the logs and observe that when the check request is published, it's always set to one of the subscriptions, and does not update based on the proxy client's region. Example logs:
```
{"timestamp":"2018-04-30T18:18:35.560147+0000","level":"info","message":"publishing check request","payload":{"command":"check-ping.rb -h 192.168.2.1","enable_deprecated_filtering":false,"proxy_requests":{"client_attributes":{"client_type":"router"}},"name":"check-router","source":"region2-router","issued":1525112315},"subscribers":["roundrobin:check-region1-network"]}
```

## Context

In environments with many proxy clients which need checks to be executed in particular environments, it is not feasible to create a new check for each different environment or region, and instead would be more streamlined if the proxy checks could use the token substitution when building the subscriber to send the check to.

## Environment
OS: CentOS 6
Sensu Server: 1.3.3

amdprophet:

## Description
Adds a new `deep_dup` method to handle recursively cloning objects with nested values. Fixes a case in which a shallow duplicated check hash was mutated causing the original check hash to also be mutated.

## Related Issue
Closes #1855 

## Motivation and Context
See #1855 

## How Has This Been Tested?
Reproduced the issue locally. The issue no longer persists after deep-duping the check data. Added specs.

## Screenshots (if appropriate):
N/A

## Types of changes

## Checklist:

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -836,7 +836,7 @@ module Sensu
           :client => client,
           :check => check
         })
-        proxy_check, unmatched_tokens = object_substitute_tokens(check.dup, client)
+        proxy_check, unmatched_tokens = object_substitute_tokens(deep_dup(check.dup), client)
         if unmatched_tokens.empty?
           proxy_check[:source] ||= client[:name]
           publish_check_request(proxy_check)
--- a/lib/sensu/utilities.rb
+++ b/lib/sensu/utilities.rb
@@ -51,6 +51,31 @@ module Sensu
       merged
     end
 
+    # Creates a deep dup of basic ruby objects with support for walking
+    # hashes and arrays.
+    #
+    # @param obj [Object]
+    # @return [obj] a dup of the original object.
+    def deep_dup(obj)
+      if obj.class == Hash
+        new_obj = obj.dup
+        new_obj.each do |key, value|
+          new_obj[deep_dup(key)] = deep_dup(value)
+        end
+        new_obj
+      elsif obj.class == Array
+        arr = []
+        obj.each do |item|
+          arr << deep_dup(item)
+        end
+        arr
+      elsif obj.class == String
+        obj.dup
+      else
+        obj
+      end
+    end
+
     # Retrieve the system hostname. If the hostname cannot be
     # determined and an error is thrown, `nil` will be returned.
     #
--- a/spec/utilities_spec.rb
+++ b/spec/utilities_spec.rb
@@ -63,6 +63,32 @@ describe "Sensu::Utilities" do
     expect(deep_merge(hash_one, hash_two)).to eq(expected)
   end
 
+  it "can deeply merge a hash" do
+    hash = {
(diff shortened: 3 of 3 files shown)


lib/sensu/server/process.rb
```diff
@@ -836,7 +836,7 @@ def publish_proxy_check_request(client, check)
           :client => client,
           :check => check
         })
-        proxy_check, unmatched_tokens = object_substitute_tokens(check.dup, client)
+        proxy_check, unmatched_tokens = object_substitute_tokens(deep_dup(check.dup), client)
```
portertech: extra dup seems excessive?

lib/sensu/utilities.rb
```diff
@@ -51,6 +51,31 @@ def deep_merge(hash_one, hash_two)
...
+    # Creates a deep dup of basic ruby objects with support for walking
+    # hashes and arrays.
+    #
+    # @param obj [Object]
+    # @return [obj] a dup of the original object.
+    def deep_dup(obj)
+      if obj.class == Hash
+        new_obj = obj.dup
+        new_obj.each do |key, value|
+          new_obj[deep_dup(key)] = deep_dup(value)
```
portertech: There's a very good chance key is usually (or always) a symbol, therefor, this will return the original object.

portertech:
idiot-train-jump

PR #1910, opened 2018-10-04
Replace proxy check result command with original definition

### LINKED ISSUE #1869: Redaction of vars on clients created from the API
When creating proxy clients through the API, passwords/sensitive vars does not seem to be handled the same way as if they're created through a regular client (in a client.json file).
## Expected Behavior

When accessing a client created via the API, vars set on that client that is subject to redaction should be redacted. Further check commands/results where token substitution is made on those vars, should also be redacted. 
## Current Behavior

Vars subject to redaction set on clients created through the API does not get redacted in any way. Check commands/results containing vars subject to redaction is not redacted in any way.
## Possible Solution

Not a technical solution, but an idea: sensitive vars should be treated the same independently of where they're created or accessed. Maybe have a scope or key through which to specify sensitive stuff and a way for clients/api to access that data, but only those clients that have subscriptions relating to the client or check that defines the sensitive object.  
## Steps to Reproduce (for bugs)

1. Create a proxy client through the API with a variable set that should trigger redaction.
2. Create a check with token substitution utilizing that var (with proxy_requests).
3. Run the check.

## Context

I'm trying to create an ipmi check where the password is needed as an option on the command line. The ipmi check is run through proxy requests and the clients are created through an api call to which the results are reported. The password is stored on the client as a hash: "ipmi": {"user": "x", "password": "y"} and is substituted in the check through :::ipmi.user::: and :::ipmi.password::: which gives me the possibility to set different passwords on different servers.
## Your Environment

* Sensu version used: 1.4.2
* Active Sensu extensions (e.g. WizardVan 0.1.0):  None
* Operating System and version (e.g. Ubuntu 14.04): CentOS 7.4
* Transport and runtime versions (e.g. RabbitMQ 3.6.2 on Erlang 18.2):  RabbitMQ 3.73, Erlang 19.3

amdprophet:

## Description
Modifies the command in check results from proxy clients with the original definition.

## Related Issue
Fixes #1869.

## Motivation and Context
The command in a check result from a client proxy can contain sensitive information post-token substitution. This change will reset the command to contain the original tokens.

## How Has This Been Tested?
Wrote a new spec.

## Types of changes

## Checklist:
- [ ] All new and existing tests passed.

### DIFF
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -703,6 +703,9 @@ module Sensu
             end
             check[:type] ||= STANDARD_CHECK_TYPE
             check[:origin] = result[:client] if check[:source]
+            if @settings.check_exists?(check[:name].to_sym) && client[:type] == "proxy"
+              check[:command] = @settings[:checks][check[:name].to_sym][:command]
+            end
             aggregate_check_result(client, check) if check[:aggregates] || check[:aggregate]
             store_check_result(client, check) do
               create_event(client, check) do |event|
--- a/lib/sensu/server/process.rb
+++ b/lib/sensu/server/process.rb
@@ -703,7 +703,7 @@ module Sensu
             end
             check[:type] ||= STANDARD_CHECK_TYPE
             check[:origin] = result[:client] if check[:source]
-            if @settings.check_exists?(check[:name].to_sym) && client[:type] == "proxy"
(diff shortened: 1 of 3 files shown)


portertech:
@amdprophet looks like you may need to do more spec setup

amdprophet:
I'm not too sure what's going on with the failing spec here. This PR doesn't touch the failing spec and the error doesn't appear to be coming from the changes I've made.

PR #1916, opened 2018-10-16
Official Docker image (CentOS 7 based)

portertech:
Inspired by @sstarcher's [Debian based Docker image](https://github.com/sstarcher/docker-sensu).

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -2,7 +2,22 @@ require 'bundler/gem_tasks'
 require 'rspec/core/rake_task'
 require 'rubocop/rake_task'
 
+require File.join(File.dirname(__FILE__), 'lib/sensu/constants')
+
 RSpec::Core::RakeTask.new(:spec)
 RuboCop::RakeTask.new
 
 task :default => :spec
+
+task :docker do
+  puts 'Creating Sensu Core CentOS Docker image ...'
+
+  release = "#{Sensu::VERSION}-1"
+
+  docker_command = "docker build -f docker/Dockerfile.centos"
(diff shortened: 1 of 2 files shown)


cjchand:
Greatness! Would love to stop building this myself, so happy to see a CentOS container might become official :)

Any thoughts on adding a docker-compose.yml to this so we can have RMQ and Redis readily available? Makes for a really nice user experience, especially for folks new to Sensu, to simply type `docker-compose up` and have a fully-working setup.

Maybe best left outside of the core effort, but maybe something to consider: I really like the construct of having a text file (e.g.: `sensu-plugins`) that you can use to bake in any required plugins at build time, ala:
`RUN for plugin in $(cat /sensu-plugins); do sensu-install -P $plugin; done`

Shane's currently does something similar at runtime, but my preference is to bake it into the image, thus speeding recovery times if/when the container dies.

Anyway, just food for thought. LMK if you want/need input or assistance on this effort. Near and dear to my ❤️

sstarcher:
@cjchand my supports it at runtime or preferably at build time when a user uses it as a base image.  I highly recommend against doing it at runtime for anything, but testing.

Another large difference that might make a difference for people is with no templating in the base container.

Any reason for CentOS as the base it's typically a very uncommon base image to use.

cjchand:
@sstarcher Ah, sorry for my confusion on the build vs run time bit.

portertech:
Merging this first pass at an official docker CentOS image. We can always create issues for desired changes and functionality.

tumblr_mvrc0evts81swhepeo1_250_zps371b1695

PR #1934, opened 2019-01-09
Add response body to health endpoint

### LINKED ISSUE #1880: Sensu API health-check should return why it fails
## Expected Behavior

Polling the Sensu API health check on `/health` should return information why the conditions failed to validate, in order to help troubleshooting issues without having to repoll the another endpoint to get more details.

## Current Behavior

Currently, polling `/health` with the right parameters may return a HTTP 412 error code which indicates the tests failed.

For example, I'm currently regularly polling `/health?messages=20&consumers=2` and I recently started to get 412 status code intermittently.

In order to understand the problems, I need to repoll `/info`, hopeing to get the same data as when `/health` initially failed. The problem was very intermittent so it was difficult to get the numbers that triggered the 412 error.

That would have been very helpful to get an answer directly in the `/health` result saying there were too much messages and/or not enough consumers, and if so, what were the limits requested and the actual numbers returned.

## Context

I'm using a in-office Sensu installation to monitor a remote Sensu installation.

The remote RabbitMQ cluster is a bit slowish right now apparently, so the Sensu servers don't process messages "fast enough". Additionnaly, all check results are coming at once, I guess at the same time as when the check which polls `/health` executes, so I guess it was actually OK to have a small spike at very specific time.

I'm tentatively willing to write a specific Sensu check to poll `/info`, move the logic which is currently in `/health` into this check and make it return the information I need in order to understand why the health logic fails when it does, but I also feel that could actually enhance the overall results from `/health` as well.

## Your Environment
* Sensu version used: 1.4.2
* Active Sensu extensions (e.g. WizardVan 0.1.0): None
* Operating System and version (e.g. Ubuntu 14.04): Debian stretch
* Transport and runtime versions (e.g. RabbitMQ 3.6.2 on Erlang 18.2): RabbitMQ 3.6.15 on Erlang 19.3

amdprophet:

## Description
Adds a new response body to the `/health` endpoint which, in the event that the health check fails, will return a list of reasons why the health check failed.

## Related Issue
Fixes #1880.

## Motivation and Context
See #1880.

## How Has This Been Tested?
rspec tests

## Types of changes

## Checklist:
- [ ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/lib/sensu/api/routes/health.rb
+++ b/lib/sensu/api/routes/health.rb
@@ -10,22 +10,32 @@ module Sensu
 
         # GET /health
         def get_health
+          @response_content = []
           if @redis.connected? && @transport.connected?
-            healthy = []
             min_consumers = integer_parameter(@params[:consumers])
             max_messages = integer_parameter(@params[:messages])
             transport_info do |info|
               if min_consumers
-                healthy << (info[:keepalives][:consumers] >= min_consumers)
-                healthy << (info[:results][:consumers] >= min_consumers)
+                if info[:keepalives][:consumers] < min_consumers
+                  @response_content << "keepalive consumers (#{info[:keepalives][:consumers]}) less than min_consumers (#{min_consumers})"
+                end
+                if info[:results][:consumers] < min_consumers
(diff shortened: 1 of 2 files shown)


portertech:
ukl8tjg

PR #1959, opened 2019-08-28
Set default pipe handler timeout

amdprophet:

## Description
Sets the default pipe handler timeout to 10 seconds.

## Motivation and Context
Sensu can indefinitely wait for child processes to quit during a clean shutdown. In the event that the parent process is forcefully killed while child processes are still running, the ports that Sensu listen on can remain "in use" preventing a restart. The only workaround is to manually kill the child processes.

Setting a default pipe handler timeout will help prevent this case from happening. Secondarily, our documentation already states that the default timeout is 10 for pipe handlers.

## How Has This Been Tested?
Tested that a pipe handler (golang app that sleeps for 60 seconds) was killed after 10 seconds of execution.

## Types of changes

## Checklist:

### DIFF
--- a/lib/sensu/server/handle.rb
+++ b/lib/sensu/server/handle.rb
@@ -39,7 +39,10 @@ module Sensu
       #   process via STDIN.
       # @param event_id [String] event UUID
       def pipe_handler(handler, event_data, event_id)
-        options = {:data => event_data, :timeout => handler[:timeout]}
+        options = {
+          :data => event_data,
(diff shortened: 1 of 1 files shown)


portertech:
This is a breaking change with significant impact.

amdprophet:
After some discussion we will not be making this change. https://github.com/sensu/sensu-docs/issues/1714 has been opened to fix the documentation.

Issue #1961, opened 2019-09-16
HTTP 204 responses can inappropriately include content


tmonk42:
HTTP 204 "No Content" should not respond with content.

The API endpoint health initializes "@response_content" to an empty array. If health is okay as determined by the array still being empty, then it calls the "no_content!" response.
https://github.com/sensu/sensu/blob/master/lib/sensu/api/routes/health.rb#L34

The def for "no_content!" appropriately sets the status code and string (204/No Content) then calls "respond".
https://github.com/sensu/sensu/blob/master/lib/sensu/api/http_handler.rb#L281

The "respond" method does a JSON.dump of "@response_content", which in the case of the health endpoint being healthy, is an empty array, which gets serialized to a response of "[]".
https://github.com/sensu/sensu/blob/master/lib/sensu/api/http_handler.rb#L205

## Expected Behavior

If response is 204/No Content, then response should not return any content (including empty arrays).

## Current Behavior

Health endpoint responds when healthy with a 204 than includes "[]" in the content.

## Possible Solution

PR incoming shortly. In the def for "no_content!" we should nil out @response_content.

## Steps to Reproduce (for bugs)

1. Call /health api endpoint, note that headers look correct, and body include []
2.
3.
4.

## Your Environment

* Sensu version used: 1.8.0
* Active Sensu extensions (e.g. WizardVan 0.1.0):
* Operating System and version (e.g. Ubuntu 14.04): Ubuntu 18.04
* Transport and runtime versions (e.g. RabbitMQ 3.6.2 on Erlang 18.2):

portertech:
Closed by #1962

PR #1962, opened 2019-09-16
HTTP response 204 'no content' should actually have no content

tmonk42:
Signed-off-by: Nathan Haneysmith <nathan@chef.io>

When preparing a HTTP 204 response, nil out content.

## Description

Currently any value in @response_content will be returned in a 204 response. This can be see in the health api endpoint, where a "good" result returns the value of @response_content, which is an empty array "[]" which then gets JSON.dump'd into the response content as "[]". This should properly be "no content" and not a json representation of an empty array.

This came up as I was attempting to setup AWS ALB host group healthcheck on the health endpoint. The non-empty response while returning status code 204 (no response) causes an error.

## Related Issue

Fix for #1961 

## Motivation and Context

This will bring us in line with RFC for HTTP 204 and allow AWS ALB Target Group healthchecks to appropriately call the /health endpoint

## Types of changes

## Checklist:

### DIFF
--- a/lib/sensu/api/http_handler.rb
+++ b/lib/sensu/api/http_handler.rb
@@ -281,6 +281,7 @@ module Sensu
       def no_content!
         @response_status = 204
         @response_status_string = "No Content"
+        @response_content = nil
         respond
       end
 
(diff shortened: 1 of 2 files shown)


tmonk42:
Travis is failing due to "Failed to start rabbitmq-server.service: Unit rabbitmq-server.service not found" at around 2 seconds into the run...

